Issue #8, opened 2026-05-30
PetBot 2.0 — tracking epic


auzbuzzard:
Tracking epic for the PetBot 2.0 revival. The 2018 bot (discord.py 0.16 "async" branch, unrunnable today) was rebuilt on **discord.py 2.x** as **ports & adapters**: a platform-neutral core (`petbot.core`, never imports `discord`) with a Discord adapter as the only frontend. Legacy snapshot preserved at tag `v0.1-legacy-2018`.

CI is intentionally **offline & secret-free** (mocked APIs, no Discord login), so anything needing a live token/guild/network is tracked as a child issue below.

## ✅ Landed on `master`
- **#7** — platform-neutral revival: core (`Skill`/`SkillContext`/`SkillResult`/`Capabilities`/`SkillRegistry`/`VoicePort`), skills (`math`, `derpi`, `e621`, `music`), Discord adapter (`/ping /math /derpi /e621 /music /purge`), tooling/CI (ruff, mypy --strict, import-linter, pytest), docs + ADRs.
- **#10** — music queue auto-advances when a track finishes (with a stale-callback guard).
- **#24** — booru providers rewritten around typed pydantic models + a shared `engine.py` (providers / `render.py` / reworked error handling).
- **#26** — modern, 3.12-native `logging` adopted across the codebase (closed the #25 stub).
- **#34** — serverless deployment groundwork: ADR `0005-serverless-deployment.md` + the **HTTP-Interactions adapter** (`petbot/frontends/interactions/`), which completed **#29**.

## 🔭 In flight
- **PR #37** — cleanup follow-ups: ports the one still-relevant fix from the now-closed draft **PR #11** (Derpibooru protocol-relative URL absolutization; the e621 null-url fix already landed via #24), repoints the deferred-path code comments at #35, and resolved the duplicate `0004` ADR number (serverless renumbered to `0005`; logging keeps `0004`).

## 📋 Remaining work (child issues)
- [ ] **#28** — **Deploy PetBot** — serverless core on AWS Lambda (+ future hybrid cluster skills) — **epic**
  - [x] **#29** — Interactions frontend adapter — *done in #34*
  - [ ] **#30** — Terraform: Lambda + Function URL + IAM
  - [ ] **#31** — Discord application + slash-command/endpoint registration *(now also owns the `auzbuzzard.net` domain decision, folded from #32)*
  - [ ] **#35** — Deferred-response mechanism + ordered multi-message delivery
[...]

PR #26, opened 2026-06-02
Adopt modern, 3.12-native logging across the codebase

### LINKED ISSUE #25: Adopt Pythonic logging across the codebase
**Stub — to be fleshed out.**

Adopt idiomatic Python `logging` across the codebase (replace ad-hoc/`print`-style output):

- module-level `logger = logging.getLogger(__name__)` per module; **no** logging config at import time
- configure handlers/levels **once** in the entrypoint (`petbot/__main__.py` / Discord bootstrap); level via env (e.g. `LOG_LEVEL`)
- sensible levels: DEBUG for request/parse internals, INFO for lifecycle, WARNING/ERROR for failures
- never log secrets/tokens

**Depends on #24** — land the booru refactor first (new `engine.py` / providers / `render.py` + reworked error handling), so logging is wired into the final structure instead of code that's about to be rewritten.

Related: #8 (2.0 tracking epic).

auzbuzzard:
## Why

Issue #25 (sequenced after the booru rewrite #24, now landed) asks for idiomatic
Python `logging` across the codebase. The repo shipped with a lone
`logging.basicConfig(level=INFO)` in `bootstrap.run` and otherwise silence. This
adopts the modern, 3.12-native shape from mCoding's [modern logging talk](https://youtu.be/9L77QExPmI0)
([sample code](https://github.com/mCodingLLC/VideosSampleCode/tree/master/videos/135_modern_logging)).

## What changed

**One config point, per-module loggers.** No module configures logging at import
time. `configure_logging()` in the new `petbot/logging_setup.py` is the single
setup point, called from `bootstrap.run`; every module does
`logger = logging.getLogger(__name__)` and just emits. `bot.run(..., log_handler=None)`
(already present) lets discord.py's `discord.*` loggers propagate into our handlers.

**`dictConfig` from JSON, shipped as package data.** Config is data, not code —
JSON loaded via `importlib.resources` + `logging.config.dictConfig`. Two profiles:

| Profile | When | Shape |
|---|---|---|
| `plain` | dev (default) | one human-readable line → stderr |
| `structured` (`json`) | prod (default) | JSON-lines, **non-blocking** via `QueueHandler`/`QueueListener`; INFO/DEBUG → stdout, WARNING+ → stderr |

The queue profile keeps log I/O off the asyncio event loop (AGENTS invariant #4).
The listener is started in `configure_logging` and stopped once via `atexit`.

**Custom `JSONFormatter` + `NonErrorFilter`** using `typing.override`,
`datetime.UTC`, and `logging.getHandlerByName` — 3.12-only APIs, which is the
concrete reason the project is firmly 3.12+ (`requires-python` already pinned;
README bumped to match).

**Env-driven, via `config.py` only.** `Settings` gains `LOG_LEVEL` (default
`INFO`) and `LOG_FORMAT` (`plain`|`json`, else derived from `ENV`), parsed in
`from_env` with a `resolved_log_format` helper.

**Instrumentation across core + adapter.** Sensible calls at DEBUG (booru engine
request/parse internals, tag building, command breadcrumbs), INFO (login/sync,
now-playing, purge), and WARNING/ERROR (network/parse failures, the previously
*silent* greeter fallback, unexpected `/purge` errors with `exc_info`). **Secrets
are never logged** — booru searches are logged from the neutral `SearchRequest`,
never the wire request (which can carry an api-key/User-Agent).

## Tests & docs

- `tests/test_logging.py` — formatter output (required fields, `fmt_keys`, merged
  extras, exception capture), `NonErrorFilter`, `configure_logging` (plain/queued,
  bad format/level), and `Settings` log parsing.
- `docs/adr/0004-logging.md` + a Logging section in `docs/architecture.md`; new
  invariant in `AGENTS.md`; `.env.example` documents the knobs.

## Verification

`ruff check` · `ruff format --check` · `mypy` (strict, 44 files) · `lint-imports`
(core/adapter boundary intact) · `pytest` — all green (63 passed, 2 live skipped),
on Python 3.12. Manually smoke-tested both profiles end to end.

Closes #25. Related: #8 (2.0 epic).

https://claude.ai/code/session_01NjVqTu47jRaTuNUHvJZm4T

---
_Generated by [Claude Code](https://claude.ai/code/session_01NjVqTu47jRaTuNUHvJZm4T)_

### DIFF
--- a/petbot/core/capabilities/boorus/engine.py
+++ b/petbot/core/capabilities/boorus/engine.py
@@ -44,27 +44,24 @@ async def run_search(
     logger.debug("%s responded HTTP %d", provider.name, response.status_code)
     body = _json_body(response)
 
+    # Each failure below is signalled by *raising* — the exception is the error
+    # report. We deliberately don't log here: that would double-log (the caller
+    # that handles the exception is the right place to decide level + traceback).
+
     # 1. The site's own error message wins (best UX) — when the body decoded.
     if body is not None and (reason := provider.error(body)) is not None:
-        logger.warning("%s rejected the search: %s", provider.name, reason)
         raise SiteFailureStatusError(
             site_message=reason,
             print_message=f"uwu I couldn't do that. {provider.name} says: {reason}",
         )
     # 2. Any other non-2xx is still an error, even with no recognizable error body.
     if response.status_code >= 400:
-        logger.warning(
-            "%s returned HTTP %d with no recognizable error body",
-            provider.name,
-            response.status_code,
-        )
         raise SiteFailureStatusError(
             site_message=f"HTTP {response.status_code}",
             print_message=f"uwu {provider.name} returned an error (HTTP {response.status_code}).",
         )
     # 3. A 2xx we couldn't decode is an anomaly — surface it, don't fake "no results".
     if body is None:
-        logger.warning("%s returned a non-JSON %d response", provider.name, response.status_code)
         raise SiteFailureStatusError(
             site_message="non-JSON response",
             print_message=f"uwu {provider.name} sent a response I couldn't read.",
--- a/petbot/core/capabilities/boorus/engine.py
+++ b/petbot/core/capabilities/boorus/engine.py
@@ -9,6 +9,8 @@ body still raises rather than silently looking like "no results".
 
 from __future__ import annotations
 
+import logging
+
 import httpx
 
 from petbot.core.capabilities.boorus.base import BooruProvider
@@ -16,6 +18,8 @@ from petbot.core.capabilities.boorus.errors import SiteFailureStatusError
 from petbot.core.capabilities.boorus.render import render
 from petbot.core.capabilities.boorus.types import SearchRequest
 from petbot.core.skills.context import SkillResult
+
+logger = logging.getLogger(__name__)
 
 
 async def run_search(
@@ -26,24 +30,41 @@ async def run_search(
     author: str,
 ) -> SkillResult:
     """Send the search, surface any site/HTTP error, then render the first result."""
+    # Logged from the neutral SearchRequest, never the wire request — the latter
+    # can carry an api_key/User-Agent, and secrets must never reach the logs.
+    logger.debug(
+        "%s search: tags=%s safe_only=%s sort=%s",
+        provider.name,
+        search.tags,
+        search.safe_only,
+        search.sort,
+    )
     request = provider.build_request(client, search)
     response = await client.send(request)
+    logger.debug("%s responded HTTP %d", provider.name, response.status_code)
     body = _json_body(response)
 
     # 1. The site's own error message wins (best UX) — when the body decoded.
     if body is not None and (reason := provider.error(body)) is not None:
+        logger.warning("%s rejected the search: %s", provider.name, reason)
         raise SiteFailureStatusError(
             site_message=reason,
             print_message=f"uwu I couldn't do that. {provider.name} says: {reason}",
         )
     # 2. Any other non-2xx is still an error, even with no recognizable error body.
     if response.status_code >= 400:
+        logger.warning(
+            "%s returned HTTP %d with no recognizable error body",
+            provider.name,
+            response.status_code,
+        )
         raise SiteFailureStatusError(
             site_message=f"HTTP {response.status_code}",
             print_message=f"uwu {provider.name} returned an error (HTTP {response.status_code}).",
         )
     # 3. A 2xx we couldn't decode is an anomaly — surface it, don't fake "no results".
     if body is None:
+        logger.warning("%s returned a non-JSON %d response", provider.name, response.status_code)
         raise SiteFailureStatusError(
             site_message="non-JSON response",
             print_message=f"uwu {provider.name} sent a response I couldn't read.",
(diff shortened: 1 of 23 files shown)


petbot/core/capabilities/boorus/engine.py
```diff
@@ -26,24 +30,41 @@ async def run_search(
...
+        search.sort,
+    )
     request = provider.build_request(client, search)
     response = await client.send(request)
+    logger.debug("%s responded HTTP %d", provider.name, response.status_code)
     body = _json_body(response)
 
     # 1. The site's own error message wins (best UX) — when the body decoded.
     if body is not None and (reason := provider.error(body)) is not None:
+        logger.warning("%s rejected the search: %s", provider.name, reason)
```
auzbuzzard: why are we logging debug? What's the correct pythonic way to log raised Error?

petbot/core/capabilities/boorus/engine.py
```diff
@@ -26,24 +30,41 @@ async def run_search(
...
             site_message=reason,
             print_message=f"uwu I couldn't do that. {provider.name} says: {reason}",
         )
     # 2. Any other non-2xx is still an error, even with no recognizable error body.
     if response.status_code >= 400:
+        logger.warning(
+            "%s returned HTTP %d with no recognizable error body",
+            provider.name,
+            response.status_code,
+        )
```
auzbuzzard: again, what's the logging vs raise

petbot/core/capabilities/boorus/engine.py
```diff
@@ -26,24 +30,41 @@ async def run_search(
...
         raise SiteFailureStatusError(
             site_message=f"HTTP {response.status_code}",
             print_message=f"uwu {provider.name} returned an error (HTTP {response.status_code}).",
         )
     # 3. A 2xx we couldn't decode is an anomaly — surface it, don't fake "no results".
     if body is None:
+        logger.warning("%s returned a non-JSON %d response", provider.name, response.status_code)
```
auzbuzzard: logging vs raise

auzbuzzard (commented):
1. explain how correct pythonic logging handles logging errors. Why are we writing warning messages again if the exception system are there?
2. audit all the info and debug and warning messages. Check if they are too verbose or not.

auzbuzzard:
Good catch — those `engine.py` warnings were a log-and-raise anti-pattern. Fixed in 1eac60e.

**1. Correct pythonic error logging**

The exception *is* the error report. Once you `raise`, the error is already propagating to whoever can deal with it — so you log it **once, at the boundary that actually handles it** (catches it and doesn't re-raise), not at the `raise` site. Logging in both places double-logs the same event and couples low-level code to a logging policy it shouldn't own. The rule of thumb:

- **Raising code logs nothing** — it raises. (So `engine.run_search` now just raises `SiteFailureStatusError`; the three `logger.warning(...)` calls are gone.)
- **Code that re-raises logs nothing** — let it propagate.
- **The handler logs**, picking the level by severity and attaching a traceback (`exc_info=True`, or the `logger.exception()` shorthand) **only when a traceback helps debugging**.

So the level/`debug` question on line 49: that DEBUG is in the *handler* (`booru_skill._run`), and it's DEBUG on purpose. A site rejecting a query (e.g. e621's 422 for too many tags) is an **expected, fully-handled** outcome that we turn into a friendly message — it's not a bug, so it's a DEBUG breadcrumb with no traceback. The *other* branch — we couldn't reach or decode the site at all — is **unexpected**, so it's a `WARNING` with `exc_info=True`. Two failure modes, two levels, one log each, all at the handler:

```python
try:
    return await run_search(...)
except SiteFailureStatusError as exc:      # expected + handled -> DEBUG, no traceback
    logger.debug("%s rejected the search: %s", provider.name, exc.site_message)
    return SkillResult.failure(exc.print_message)
except (httpx.HTTPError, ValueError):      # unexpected -> WARNING + traceback
    logger.warning("%s search failed to reach/parse the site", provider.name, exc_info=True)
    return SkillResult.failure(_NETWORK_FAILURE)
```

**2. Verbosity audit** (with the above as the yardstick — what shows at the default `INFO`?)

- **engine** — removed the 3 log-and-raise WARNINGs. Kept the two DEBUG breadcrumbs (request params, response status) — that's the "DEBUG for request/parse internals" the issue called for, and they're off by default.
- **math** — dropped the happy-path DEBUG (`expr -> result`); the result already goes to the user, so there's nothing to log. Kept the could-not-evaluate DEBUG.
- **music** — downgraded `stopped playback` from INFO to DEBUG; `now playing` stays the single INFO lifecycle event.
- **Kept as-is:** startup (INFO ×3, once each), `purge` (INFO, audit trail), the greeter-fallback WARNING (only fires on a broken resource), and the cog `/cmd invoked by <user>` DEBUGs — those live in the adapter and log *who invoked what*, which the neutral core deliberately doesn't see, vs. the core's DEBUG which logs the neutral operation. Different layers, both DEBUG.

Net effect at default `INFO`: lifecycle + genuine failures only; everything chatty is DEBUG. ADR 0004 now spells out the log-once-at-the-handler principle. Full gate suite green on 3.12 (ruff/format/mypy/lint-imports/pytest).

---

Issue #31, opened 2026-06-04
Discord application + slash-command registration + interactions endpoint URL


auzbuzzard:
Sub-issue of #28. Create/recover the Discord app(s), register commands, and point Discord at the deployed Lambda — all scripted and repo-owned where the API allows.

## Bot/environment model
- **Dev** → app "PetBot Dev", `ENV=dev`, commands registered **guild-scoped** to the dev guild (instant). Invited to the dev guild (`qJXBSt8wq`).
- **Prod/UAT** → app "PetBot Prod", `ENV=prod`, commands registered **globally**. Invited to the main server (`fwFx36mTR`).
- Recover existing apps if they still exist; otherwise create fresh and **Reset Token**.

## Tasks
- [ ] Confirm intents: **no privileged intents** needed for the current feature set; leave **Message Content OFF** (only the future Phase B chat layer #15 needs it).
- [ ] OAuth2 invite URLs — scopes `bot` + `applications.commands`; perms: Send Messages, Embed Links, Read Message History, **Manage Messages** (gates `/purge`), Connect/Speak (for future music). Invite each bot to its server.
- [ ] **`register.py`** — `PUT /applications/{id}/commands` (guild-scoped for dev, global for prod) from the skills' `input_schema`. Run in CI / on deploy.
- [ ] **Set the Interactions Endpoint URL** — `PATCH /applications/@me` with the #30 Function URL (or the custom domain decided below). **Ordering:** Lambda must be live first — Discord validates by sending a `PING` and only accepts the URL if the handler answers. Script as a post-`apply` step (`local-exec` / CI job).
- [ ] Capture `DISCORD_PUBLIC_KEY` + token into the secret store used by #30.

## Domain / endpoint hostname (folded in from #32)
Decide **whether and how** PetBot uses the `auzbuzzard.net` domain (DNS managed by Cloudflare) for its interactions endpoint, and align AWS ↔ Cloudflare. Discord accepts **any** valid HTTPS URL, so the raw **Lambda Function URL** works with zero domain wiring; a custom domain (e.g. `petbot.auzbuzzard.net`) is a *nice-to-have* for stability/branding, but **Lambda Function URLs don't natively support custom domains**, so it needs deliberate plumbing.

Options to weigh:
1. **No custom domain** — use the Function URL directly. Simplest; the URL changes only if the function is recreated. **Leaning this for the initial cut**; revisit at steady-state.
[...]

Issue #30, opened 2026-06-04
Terraform: AWS Lambda + Function URL + IAM (repo-owned deploy)


auzbuzzard:
Sub-issue of #28. Infrastructure-as-code so the repo owns the deployment end-to-end — no console click-ops.

## Scope (`deploy/terraform/`)
- `aws_lambda_function` running the #29 interactions handler (Python runtime; Graviton/arm64 for cost). *(The handler itself landed in #34.)*
- **`aws_lambda_function_url`** — public HTTPS entry (no API Gateway needed); Discord POSTs straight to it. `AUTH_NONE` (the Lambda verifies Discord's Ed25519 signature itself).
- `aws_iam_role` + least-privilege policy for the function.
- Secrets/config wiring: `DISCORD_PUBLIC_KEY`, `DISCORD_TOKEN`, booru auth, `ENV` — via SSM Parameter Store / Secrets Manager (or resolved from 1Password into the build/deploy env). **No secrets in state or repo.**
- Outputs: the Function URL — consumed by #31 (to set Discord's interactions endpoint, and for the optional custom-domain decision now tracked there).
- **Stubbed/commented future blocks** for the music bus: `aws_sqs_queue` + consumer IAM (activated by #33), plus a scoped `lambda:InvokeFunction`-on-itself permission for the deferred-response self-invoke (#35).

## Notes
- `terraform apply` should stand the whole thing up; teardown via `terraform destroy`.
- Packaging: container image or zip; keep cold-start lean.
- CI: a workflow runs `fmt`/`validate`/`plan` on PR and `apply` on merge (state in S3 + DynamoDB lock, or TF Cloud — decide here).

## Done when
`terraform apply` produces a live Function URL serving the #29 handler, with all config supplied as code/secrets references.

auzbuzzard:
## Runbook notes (from the planning session)

- **Packaging → container image.** The adapter pulls in C-extensions (`pynacl` for Ed25519, `numexpr` for `/math`), so a **Lambda container image** (reusing the existing `Dockerfile` pattern, on arm64/Graviton) is the reliable choice — it dodges manylinux-wheel wrangling that a zip build invites. Keep the image lean for cold starts.
[...]

Issue #29, opened 2026-06-04
Interactions frontend adapter (HTTP Interactions over the neutral core)


auzbuzzard:
Sub-issue of #28. The app-code prerequisite for the serverless deployment.

Add a new transport adapter under **`petbot/frontends/interactions/`** that serves Discord's **HTTP Interactions** model (Discord POSTs each slash command to a URL) and reuses the existing neutral `SkillRegistry` — the stateless counterpart to the gateway adapter in `petbot/frontends/discord/`.

## Scope
- **`verify.py`** — Ed25519 signature check (`X-Signature-Ed25519` / `X-Signature-Timestamp` against the app public key); answer the `PING` (type 1) with `PONG`. Discord rejects the endpoint at registration time if this is wrong.
- **`context.py`** — map the interaction JSON → neutral `SkillContext`. Set `Capabilities(supports_voice=False, supports_rich_embeds=True, max_text_length=2000)`; derive `allows_explicit` from the interaction's channel `nsfw` flag.
- **`render.py`** — `EmbedSpec` → **interaction-response JSON** (raw dicts, **not** `discord.Embed`), so this adapter pulls in **no `discord.py`** and stays light/portable.
- **`app.py`** — ASGI/handler entrypoint (FastAPI or a thin handler). Build the registry with the **5 stateless skills** (`ping`, `math`, `derpi`, `e621`, `purge`). Because `supports_voice=False`, the registry **auto-hides `music`** — no special-casing. Compose deps (httpx client, booru auth) at this root.
- **Deferred responses** — boorus can exceed Discord's 3s limit. Default to answering inline (boorus are usually <3s); leave a documented hook for true deferral (ACK type 5 → follow-up PATCH) if a call exceeds 3s.

## Invariants
- Must **not** be imported by `petbot.core` (enforced by `import-linter` + `tests/test_core_isolation.py`). Keep it `discord`-free too, so it runs on minimal runtimes.
- Adding this adapter requires **zero changes** to existing skills.
- Tests offline (fake signature keypair + sample interaction payloads), consistent with `AGENTS.md`.

## Optional task (host-portability)
- [ ] Swap `/math` from `numexpr` to a small pure-Python AST evaluator. **Not needed on Lambda** (native Python keeps `numexpr`); only required to keep **Cloudflare Workers** open as an alternative host. Also drops a heavy native dep.

## Done when
[...]

Issue #28, opened 2026-06-04
Epic: Deploy PetBot — serverless core on AWS Lambda (+ future hybrid cluster skills)


auzbuzzard:
**Epic.** Revive PetBot's live deployment using the serverless split we settled on, getting a running bot onto the **dev guild** (dev environment) and the **main server** (prod + UAT). Extends/supersedes the gateway-container plan in #17; part of the 2.0 epic #8.

## Decision (summary)
PetBot splits into **transport adapters over the neutral core**, deployed as independent units that rendezvous at runtime:

- **Stateless skills** (`/ping /math /derpi /e621 /purge`) → **HTTP Interactions on AWS Lambda** (Function URL). Free-tier (~$0), public HTTPS, **no inbound exposure to home**. Infra owned by **Terraform**, in-repo.
- **Gateway-required skills** (`/music`, future chat/LLM) → a **persistent worker image on the homelab cluster**, reached from Lambda via a **message-bus queue** as a `RemoteSkill` proxy — so the Lambda core is **agnostic to where each skill runs** (in-process vs. remote container). The worker *pulls* jobs, so still **no inbound to home**.
- **Topology: Option C** — one Discord application; all slash commands via HTTP interactions; music as a backend worker fed by a queue.
- **Music is deferred** for the initial cut → near-term deployable is a **single Lambda**. Re-activating music later is purely additive (a `RemoteMusicSkill` proxy + the cluster worker + the queue), with **no change** to the other skills.

Rationale (sourced): Cloudflare Python Workers are viable now (httpx works) but still beta and lack outbound WebSocket/UDP, so voice can never run there; Lambda keeps `numexpr` and the existing Python skills unchanged; serverless front-door is $0 on Workers/Lambda/DO at this scale; voice is outbound-only (no tunnel, no TOS issue). Full write-up landed as ADR `0005-serverless-deployment.md` (via #34).

## Sub-issues
- [x] **#29** — Interactions frontend adapter (`petbot/frontends/interactions/`) — **done in #34**
- [ ] **#38** — Decouple config: make `DISCORD_TOKEN` gateway-only (Lambda needs only `DISCORD_PUBLIC_KEY`) — *prerequisite for #30*
- [ ] **#30** — Terraform: Lambda + Function URL + IAM
- [ ] **#42** — CI deploy pipeline (GitHub Actions + OIDC: build/push image + `terraform apply`, no local runs) — *supersedes the lint-only Terraform CI from #30*
[...]

PR #34, opened 2026-06-04
Serverless deployment: ADR 0004 + HTTP-Interactions adapter

auzbuzzard:
Records the serverless deployment decision and lands the first slice of it. Part of epic #28.

### ADR 0004 — serverless deployment topology (#28)
`docs/adr/0004-serverless-deployment.md` records the decision worked out in the deployment discussion: deploy the **stateless skills as an HTTP-Interactions adapter on AWS Lambda** (Terraform-owned), while **gateway-required skills (`/music`, future LLM) run as remote workers** reached via a message bus — keeping the Lambda core agnostic to where each skill runs. Captures the Option-C topology, the remote-skill proxy seam, the no-inbound GitOps deploy, the Lambda-vs-Workers rationale, and that `/music` is deferred for the first cut.

### Interactions frontend adapter (#29)
New `petbot/frontends/interactions/` — the stateless counterpart to the gateway adapter, over the **same neutral `SkillRegistry`**:

| Module | Role |
| --- | --- |
| `verify.py` | Ed25519 request-signature check (PyNaCl) |
| `context.py` | interaction JSON → neutral `SkillContext` (`allows_explicit` from the channel NSFW flag; `supports_voice=False`) |
| `render.py` | `EmbedSpec`/`SkillResult` → interaction-response JSON (**no `discord.py`**) |
| `handler.py` | verify → parse → dispatch → render; PING/PONG, `/ping`, and the registry skills |
| `app.py` | composition root + AWS Lambda Function URL entrypoint |

Because the adapter declares `supports_voice=False`, the registry **auto-hides `/music`** — no special-casing (a behavioural test asserts this).

## Notable points
- **Imports no `discord`** (emits raw JSON) so it runs on minimal serverless runtimes; a **new import-linter contract enforces this**.
- Adds `Settings.discord_public_key` (+ `.env.example`), a `pynacl` **`interactions` optional-dependency** (kept out of the base install), and an `nacl.*` mypy override.
- **Offline tests** use a real Ed25519 keypair generated per test — the verification path is genuinely exercised, not stubbed. No network.
- Zero changes to existing skills or the gateway adapter.

## Known gaps (tracked, intentional)
- **`/purge`** acts on Discord directly (REST bulk-delete + Manage Messages gating) rather than via a neutral skill, so it's not handled here yet — a follow-up under #28.
- Multi-message output (text > 2000 chars) keeps only the first chunk; the deferred-follow-up path is a documented TODO (not needed by the current stateless skills).
- The Lambda binding (`lambda_handler`) ships here; the **Terraform** that points a Function URL at it is #30.

## Gates
Local run on Python 3.12: `ruff check` ✓ · `ruff format --check` ✓ · `mypy --strict` ✓ · `lint-imports` ✓ (3 contracts kept, incl. the new one) · `pytest` ✓ (66 passed, 2 skipped; 16 new tests).

## Next
Build order from here: **#30 Terraform** → **#31 Discord registration** → **#14 smoke test**.

Refs #28, #29.

---
_Generated by [Claude Code](https://claude.ai/code/session_015MvJMSRRpchpW1PB7jHDjA)_

### DIFF
--- a/petbot/frontends/interactions/app.py
+++ b/petbot/frontends/interactions/app.py
@@ -1,0 +1,91 @@
+"""Composition root and AWS Lambda Function URL entrypoint.
+
+:func:`build_handler` wires the stateless skills into an
+:class:`InteractionHandler`; :func:`lambda_handler` is the AWS Lambda Function
+URL binding. The Terraform stack (#30) points the function at
+``petbot.frontends.interactions.app.lambda_handler``.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import base64
+import json
+from collections.abc import Mapping
+from typing import Any
+
+import httpx
+
+from petbot.config import ConfigError, Settings
+from petbot.core.skills.booru_skill import DerpiSkill, E621Skill
+from petbot.core.skills.context import Capabilities
+from petbot.core.skills.math_skill import MathSkill
+from petbot.core.skills.registry import SkillRegistry
+from petbot.frontends.interactions.handler import InteractionHandler
+
+#: Capabilities of the stateless HTTP-interactions frontend: no voice (so the
+#: registry hides ``/music``), rich embeds, Discord's 2000-char limit.
+INTERACTIONS_CAPABILITIES = Capabilities(
+    supports_voice=False,
+    supports_rich_embeds=True,
+    max_text_length=2000,
+)
+
+
+def build_handler(settings: Settings, *, http_client: httpx.AsyncClient) -> InteractionHandler:
+    """Wire the stateless skills into an :class:`InteractionHandler`.
+
+    Raises :class:`ConfigError` if ``DISCORD_PUBLIC_KEY`` is unset — it is
+    required to verify Discord's request signatures.
+    """
+    if not settings.discord_public_key:
+        raise ConfigError(
+            "DISCORD_PUBLIC_KEY is not set. The HTTP-Interactions frontend needs the "
+            "application's public key (Discord Developer Portal → General Information) "
+            "to verify request signatures."
+        )
+    registry = SkillRegistry(
+        [
+            MathSkill(),
+            DerpiSkill(client=http_client, api_key=settings.derpibooru_api_key),
+            E621Skill(
+                client=http_client,
+                user_agent=settings.user_agent,
+                username=settings.e621_username,
+                api_key=settings.e621_api_key,
+            ),
+        ]
+    )
+    return InteractionHandler(
+        registry=registry,
+        capabilities=INTERACTIONS_CAPABILITIES,
+        public_key=settings.discord_public_key,
+    )
+
+
+def _extract(event: Mapping[str, Any]) -> tuple[bytes, str | None, str | None]:
+    """Pull the raw body and signature headers out of a Function URL event."""
+    headers = {k.lower(): v for k, v in (event.get("headers") or {}).items()}
+    raw = event.get("body") or ""
+    body = base64.b64decode(raw) if event.get("isBase64Encoded") else raw.encode()
+    return body, headers.get("x-signature-ed25519"), headers.get("x-signature-timestamp")
+
+
+async def _ahandle(event: Mapping[str, Any]) -> tuple[int, dict[str, Any]]:
+    settings = Settings.from_env()
+    # A short-lived client per invocation keeps it bound to this event loop. A
+    # warm-reuse optimization can come later if traffic warrants it.
+    async with httpx.AsyncClient(timeout=20.0) as client:
+        handler = build_handler(settings, http_client=client)
+        body, signature, timestamp = _extract(event)
+        return await handler.handle(body, signature=signature, timestamp=timestamp)
+
+
+def lambda_handler(event: Mapping[str, Any], context: Any = None) -> dict[str, Any]:
+    """AWS Lambda Function URL handler: parse the event, run, return a response."""
+    status, payload = asyncio.run(_ahandle(event))
+    return {
+        "statusCode": status,
+        "headers": {"content-type": "application/json"},
+        "body": json.dumps(payload),
+    }
--- a/petbot/frontends/interactions/app.py
+++ b/petbot/frontends/interactions/app.py
@@ -41,7 +41,7 @@ def build_handler(settings: Settings, *, http_client: httpx.AsyncClient) -> Inte
     if not settings.discord_public_key:
         raise ConfigError(
             "DISCORD_PUBLIC_KEY is not set. The HTTP-Interactions frontend needs the "
-            "application's public key (Discord Developer Portal → General Information) "
+            "application's public key (Discord Developer Portal, General Information) "
             "to verify request signatures."
         )
     registry = SkillRegistry(
--- a/petbot/frontends/interactions/app.py
+++ b/petbot/frontends/interactions/app.py
@@ -22,6 +22,7 @@ from petbot.core.skills.context import Capabilities
 from petbot.core.skills.math_skill import MathSkill
 from petbot.core.skills.registry import SkillRegistry
 from petbot.frontends.interactions.handler import InteractionHandler
+from petbot.logging_setup import configure_logging
 
 #: Capabilities of the stateless HTTP-interactions frontend: no voice (so the
 #: registry hides ``/music``), rich embeds, Discord's 2000-char limit.
@@ -71,8 +72,26 @@ def _extract(event: Mapping[str, Any]) -> tuple[bytes, str | None, str | None]:
     return body, headers.get("x-signature-ed25519"), headers.get("x-signature-timestamp")
 
 
-async def _ahandle(event: Mapping[str, Any]) -> tuple[int, dict[str, Any]]:
-    settings = Settings.from_env()
+#: Set once per warm container so :func:`configure_logging` (which restarts the
+#: queue listener on each call) runs a single time, not per invocation.
+_logging_configured = False
+
+
+def _ensure_logging_configured(settings: Settings) -> None:
+    """Configure process logging once per container (never at import time).
+
+    Mirrors the gateway, which configures logging in ``bootstrap.run``; the
+    Lambda entrypoint is the equivalent start-up point here, so the interactions
+    frontend emits the same structured/plain logs (see ADR 0004) instead of the
+    root logger's unconfigured default.
+    """
+    global _logging_configured
+    if not _logging_configured:
+        configure_logging(level=settings.log_level, fmt=settings.resolved_log_format)
+        _logging_configured = True
+
+
+async def _ahandle(event: Mapping[str, Any], settings: Settings) -> tuple[int, dict[str, Any]]:
     # A short-lived client per invocation keeps it bound to this event loop. A
     # warm-reuse optimization can come later if traffic warrants it.
     async with httpx.AsyncClient(timeout=20.0) as client:
@@ -83,7 +102,9 @@ async def _ahandle(event: Mapping[str, Any]) -> tuple[int, dict[str, Any]]:
 
 def lambda_handler(event: Mapping[str, Any], context: Any = None) -> dict[str, Any]:
     """AWS Lambda Function URL handler: parse the event, run, return a response."""
-    status, payload = asyncio.run(_ahandle(event))
+    settings = Settings.from_env()
+    _ensure_logging_configured(settings)
+    status, payload = asyncio.run(_ahandle(event, settings))
     return {
         "statusCode": status,
         "headers": {"content-type": "application/json"},
--- a/petbot/frontends/interactions/handler.py
+++ b/petbot/frontends/interactions/handler.py
@@ -1,4 +1,4 @@
-"""The interaction handler: verify → parse → dispatch → render.
+"""The interaction handler: verify -> parse -> dispatch -> render.
 
 Framework-agnostic on purpose. :meth:`InteractionHandler.handle` takes the raw
 request body plus the two signature headers and returns an
@@ -8,6 +8,7 @@ HTTP server and bound to a concrete runtime (AWS Lambda) in :mod:`.app`.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 from collections.abc import Mapping
@@ -20,7 +21,13 @@ from petbot.frontends.interactions.render import message_response
 from petbot.frontends.interactions.verify import verify_signature
 from petbot.frontends.interactions.wire import APPLICATION_COMMAND, PING, PONG
 
-log = logging.getLogger("petbot.interactions")
+logger = logging.getLogger(__name__)
+
+#: Per-skill budget for the *immediate* response. Discord cancels an interaction
+#: not answered within 3 seconds of its POST; we cap a skill below that so a slow
+#: or hung skill yields a clear message instead of a silent Discord timeout. The
+#: real fix for genuinely long work is the deferred follow-up path (tracked: #33).
+DEFAULT_SKILL_TIMEOUT_SECONDS = 2.5
 
 
 class InteractionHandler:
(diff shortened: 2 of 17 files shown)


petbot/frontends/interactions/app.py
```diff
@@ -0,0 +1,91 @@
...
+        public_key=settings.discord_public_key,
+    )
+
+
+def _extract(event: Mapping[str, Any]) -> tuple[bytes, str | None, str | None]:
+    """Pull the raw body and signature headers out of a Function URL event."""
+    headers = {k.lower(): v for k, v in (event.get("headers") or {}).items()}
+    raw = event.get("body") or ""
+    body = base64.b64decode(raw) if event.get("isBase64Encoded") else raw.encode()
+    return body, headers.get("x-signature-ed25519"), headers.get("x-signature-timestamp")
```
auzbuzzard, line 72: What are we extracting here? Why do we have to do our own extraction of http headers? Do httpx or any libraries not offer this?

petbot/frontends/interactions/handler.py
```diff
@@ -0,0 +1,101 @@
...
+from typing import Any
+
+from petbot.core.skills.context import Capabilities, SkillResult
+from petbot.core.skills.registry import SkillNotFoundError, SkillRegistry
+from petbot.frontends.interactions.context import build_context
+from petbot.frontends.interactions.render import message_response
+from petbot.frontends.interactions.verify import verify_signature
+from petbot.frontends.interactions.wire import APPLICATION_COMMAND, PING, PONG
+
+log = logging.getLogger("petbot.interactions")
```
auzbuzzard: rebase this. Logging should reflect on issue #25. Redo any logging in this PR wrt that.

petbot/frontends/interactions/handler.py
```diff
@@ -0,0 +1,101 @@
...
+        itype = interaction.get("type")
+        if itype == PING:
+            return 200, {"type": PONG}
+        if itype == APPLICATION_COMMAND:
+            return 200, await self._dispatch(interaction)
+        return 200, message_response(SkillResult.failure("Unsupported interaction type."))
+
+    async def _dispatch(self, interaction: Mapping[str, Any]) -> dict[str, Any]:
+        data = interaction.get("data") or {}
+        name = str(data.get("name", ""))
```
auzbuzzard, line 86: Is repo using pylance strict?

petbot/frontends/interactions/handler.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # ``/ping`` is an adapter-level liveness check, not a neutral skill
+        # (mirrors the gateway adapter's PingCog).
+        if name == "ping":
+            return message_response(SkillResult.message("🏓 Pong!"))
```
auzbuzzard: Don't use Emoji's when unwarrented. Review all strings.

petbot/frontends/interactions/handler.py
```diff
@@ -0,0 +1,101 @@
...
+            option["name"]: option["value"]
+            for option in data.get("options", [])
+            if isinstance(option, Mapping) and "value" in option
+        }
+        ctx = build_context(interaction)
+        try:
+            skill = self._registry.get(name)
+        except SkillNotFoundError:
+            return message_response(SkillResult.failure(f"Unknown command: `/{name}`."))
+        result = await skill.run(args, ctx)
```
auzbuzzard: Should we be try guarding this?

petbot/frontends/interactions/handler.py
```diff
@@ -0,0 +1,101 @@
...
+            return 400, {"error": "invalid JSON body"}
+
+        itype = interaction.get("type")
+        if itype == PING:
+            return 200, {"type": PONG}
+        if itype == APPLICATION_COMMAND:
+            return 200, await self._dispatch(interaction)
+        return 200, message_response(SkillResult.failure("Unsupported interaction type."))
+
+    async def _dispatch(self, interaction: Mapping[str, Any]) -> dict[str, Any]:
```
auzbuzzard, line 84: Can we have a better visibility and handling of the time limit here? Discord has a 3sec limit on lambda, but we shouldn't assume skills will match this. What's the solution here?

petbot/frontends/interactions/render.py
```diff
@@ -0,0 +1,92 @@
...
+    if spec.image_url:
+        embed["image"] = {"url": spec.image_url}
+    if spec.author_name:
+        author: dict[str, Any] = {"name": spec.author_name}
+        if spec.author_url:
+            author["url"] = spec.author_url
+        if spec.author_icon_url:
+            author["icon_url"] = spec.author_icon_url
+        embed["author"] = author
+    return embed
```
auzbuzzard: Code style question. Is there a better way to do this? per line dict key assignment is not really scalable or trackable. Reviewer who read this don't know what the in and out structure is. Still pythonic, but any better, cleaner, clearer, more elegant way?

petbot/frontends/interactions/render.py
```diff
@@ -0,0 +1,92 @@
...
+            chunks.append(line[:limit])
+            line = line[limit:]
+        if len(current) + len(line) > limit:
+            chunks.append(current)
+            current = line
+        else:
+            current += line
+    if current:
+        chunks.append(current)
+    return [chunk for chunk in chunks if chunk]
```
auzbuzzard: 1. Why is this duplicated from frontend/discord? Better to have a utils class? Is that pythonic? Chunking text seems to be dependency pure. 
2. Why do we have to implement chunking ourselves? Better way? Cleaner impl?

petbot/frontends/interactions/render.py
```diff
@@ -0,0 +1,92 @@
...
+    if result.is_error:
+        return {"content": result.error}
+
+    data: dict[str, Any] = {}
+    chunks = chunk_text(result.text or "")
+    if chunks:
+        data["content"] = chunks[0]
+    if result.embed is not None:
+        data["embeds"] = [to_embed_dict(result.embed)]
+    return data
```
auzbuzzard, line 88: Why are we dropping other chunks? While current usage would not exceed limit, I don't like leaving these broken assumptions in the code. In fact, Identify and prevent all these silent assumptions and quite bug skips. This is major red flags for me.

auzbuzzard (commented):
Also include the internal review comment you have made in response to this review.

petbot/frontends/interactions/app.py
auzbuzzard: `_extract` isn't parsing an HTTP *response* — it parses the inbound **AWS Lambda Function URL event dict**: the raw request body (base64-decoded when `isBase64Encoded` is set) plus the two Ed25519 headers Discord sends (`X-Signature-Ed25519`, `X-Signature-Timestamp`). `httpx` is our *outbound* client; it has no role in parsing an inbound AWS event, and the stdlib has nothing for the AWS event shape either.

The one library that *does* model this is `aws-lambda-powertools` — `LambdaFunctionUrlEvent` gives `.decoded_body` (handles the base64) and case-insensitive `.headers`, which would replace these 3 lines.

For now it's kept hand-rolled: 3 well-tested lines vs. a new dependency + cold-start import weight to save them. The stronger case for Powertools is holistic (its structured Logger for #25, Tracer/Metrics for #30 observability), at which point the event parser comes along for free. So adopting it is tracked to land *with* #25/#30 rather than now just for header extraction. Still open to your call if you'd rather adopt it on this PR.

---

petbot/frontends/interactions/handler.py
auzbuzzard: Reworked in `d78cc06`. Handler logging is now warning/exception-level only (unsupported interaction type, unhandled command, skill timeout, unexpected skill error) — no chatty info logging. Full alignment with #25's structured-logging standard is tracked there: I'll converge this adapter's logging onto that format (and evaluate Powertools' `Logger` as the carrier) when #25 lands, rather than inventing a one-off scheme here now.

---

petbot/frontends/interactions/handler.py
auzbuzzard: The repo's type gate is **mypy `--strict`** (`[tool.mypy] strict = true` in `pyproject.toml`), run in CI — that's the source of truth, not Pylance/Pyright, which is editor-only and not a gate. Line 86's `str(data.get("name", ""))` is the coercion that satisfies strict mode: `data` is `Mapping[str, Any]`, so `.get(...)` returns `Any`; wrapping in `str(...)` pins it to `str` for the `name == "ping"` / `name not in self._allowed` checks below.

---

petbot/frontends/interactions/handler.py
auzbuzzard: Done in `d78cc06`. Emojis removed and all user-facing strings reviewed — a scan of the `interactions` package now finds none. The strings are plain text ("Pong!", "Something went wrong running that command.", etc.).

---

petbot/frontends/interactions/handler.py
auzbuzzard: Added in `d78cc06`. The skill call is now guarded on both axes: `asyncio.wait_for(skill.run(...), self._skill_timeout)` with `except TimeoutError` (explicit "took too long" response + a warning log) and a catch-all `except Exception` (logs via `logger.exception` and returns a generic failure message). So a skill that raises or hangs yields a clean interaction response rather than a 500 / silent Discord timeout.

---

petbot/frontends/interactions/handler.py
auzbuzzard: Addressed in `d78cc06`. Discord's 3s deadline and the per-skill budget are now decoupled and explicit: `DEFAULT_SKILL_TIMEOUT_SECONDS = 2.5` (constructor-injectable, documented as deliberately sitting *below* Discord's 3s POST deadline) bounds every skill via `asyncio.wait_for`. We don't assume skills match Discord's limit — on overrun the user gets an explicit "took too long" message instead of a silent Discord timeout, and we log a warning.

The real solution for genuinely long-running work is the deferred follow-up path (ACK within 3s, then edit the message via webhook) — that's tracked in #33 and intentionally out of scope for this slice.

---

petbot/frontends/interactions/render.py
auzbuzzard: Reworked in `d78cc06`. `to_embed_dict` is now a single declarative literal, so the input→output mapping is visible at a glance, with `None`-valued keys pruned at the end (Discord omits absent fields, and the nested `author` object is built the same way). No more incremental per-line `embed["x"] = ...` accretion that a reviewer has to mentally assemble.

---

petbot/frontends/interactions/render.py
auzbuzzard: Both points addressed in `d78cc06`.

1. **Duplication** — the chunker was dependency-pure (operates on `str`, no platform types), so it moved to `petbot/core/text.py::chunk_text` and is now *shared*: `frontends/discord/render.py` imports it instead of carrying its own copy. A plain function in a core module is the Pythonic home for a platform-neutral helper here (no need for a class).

2. **Implementing it ourselves** — this adapter no longer chunks at all. A single *immediate* interaction response is exactly one message in Discord's model, so there's nothing to chunk on the immediate path; it `_truncate_to_single_message` instead (see the dropped-chunks thread). `chunk_text` is the shared multi-message primitive, used where multiple sends are actually possible — the gateway adapter today, and the deferred follow-up path (#33) later.

---

petbot/frontends/interactions/render.py
auzbuzzard: Fixed in `d78cc06` — nothing is silently dropped now. Overflow past one message is kept *up to the limit* with a **visible** notice ("... (output truncated; N more characters)") and a logged warning; the notice width is reserved up front so the result is guaranteed to fit. An empty success result likewise gets an explicit placeholder + warning instead of serialising to an empty message Discord would reject.

I also swept the adapter for the other silent assumptions you flagged: the option-parsing dict-comprehension and the unhandled-command branch now carry comments stating the skip is *intentional* (flat options only / `/purge` is a tracked follow-up), not an unexamined assumption.

The remaining gap — delivering long output with *no* truncation — needs the deferred follow-up path and is tracked in #33; the difference is it's now explicit and observable rather than a hidden `chunks[0]`.

---

auzbuzzard:
## Internal review — in response to your review ([#pullrequestreview-4444799172](https://github.com/auzbuzzard/petbot/pull/34#pullrequestreview-4444799172))

Posting the internal review I'd kept in our working notes, with the open questions I'd raised now resolved by the decisions we reached. Per-thread replies are inline above; this is the consolidated record.

Most of your comments landed in commit `d78cc06` ("Address PR review: explicit limits, dedupe chunker, logging, no emoji"). The questions I'd flagged and how each was settled:

1. **Deferral semantics** — *Should an over-budget skill fail explicitly now, or should I wire the deferred follow-up immediately?* → **Explicit now, deferral later.** A `2.5s` budget (below Discord's 3s deadline) yields an explicit "took too long" message; the ACK-then-webhook follow-up is tracked in **#33**. (`handler.py:84`)

2. **Header extraction** — *Hand-roll the Function URL event parsing, or adopt `aws-lambda-powertools`?* → **Kept hand-rolled for now**, with Powertools tracked to be adopted holistically alongside #25 (structured Logger) and #30 (observability), so the dependency arrives with a multi-purpose need rather than to save 3 lines. **Still open to your call** to pull it onto this PR instead. (`app.py:71`)

3. **Render refactor** — *Per-line dict assembly vs. declarative; duplicate the chunker vs. extract it?* → **Declarative `to_embed_dict`** + **`chunk_text` extracted to `petbot/core/text.py`** and shared with the gateway adapter. (`render.py`)

4. **Silent truncation** — *Leave the single-message assumption as a known gap, or make it explicit?* → **Made explicit**: visible truncation notice + warning log, no silent `chunks[0]` drop; full multi-message delivery tracked in #33. (`render.py:88`)

5. **Logging** — handler logging trimmed to warning/exception level in `d78cc06`; full convergence onto **#25**'s standard deferred to that issue.

Type gate, for the record: the repo enforces **mypy `--strict`** in CI (`[tool.mypy] strict = true`), not Pylance/Pyright (editor-only).

The one genuinely open item is **#2** (Powertools now vs. with #25/#30). Everything else is resolved in `d78cc06`; happy to have you resolve the threads as you confirm.

---

petbot/frontends/interactions/handler.py
auzbuzzard: Correction to my reply above — it was based on a stale local checkout. After syncing to the branch head, the logging foundation from #25 (ADR 0004, `logging_setup.configure_logging`, the plain/json profiles) is **already on this branch**, so "deferred to #25" was wrong.

The actual gap: `configure_logging()` configures the **root logger** once at start-up (which is why module-level `logging.getLogger(__name__)` here is correct — it propagates to it). The gateway calls it in `bootstrap.run`, but the **Lambda entrypoint never did**, so the interactions frontend was emitting unconfigured logs.

Fixed in `869d41d` — `lambda_handler` now calls `configure_logging(level=..., fmt=...)` once per warm container (guarded so the queue listener isn't restarted per invocation; not at import time, per the setup's contract), mirroring the gateway. `mypy --strict`, `ruff`, `lint-imports` (3 kept), and `pytest` (87 passed) all green; `app.py` stays `discord`-free.

---

petbot/frontends/interactions/render.py
auzbuzzard: Why is no truncation tracked in #33? This is a separate issue. I imagine Discord bots should be able to send multiple messages in order so that the full text can be sent, no? If that is a simple change, do it now, otherwise track it in a separate issue?

petbot/frontends/interactions/render.py
auzbuzzard: No, that's not what I'm asking for at all. If the output is 10k char long, I want it to be split across multiple messages seamlessly.

petbot/frontends/interactions/render.py
auzbuzzard: Agreed this is a separate concern from #33, and you're right that the goal is full text across ordered messages, not truncation. I'd also walked back the "single Lambda invocation" framing: for an HTTP-interactions bot the immediate response *is* the Lambda's HTTP return, and returning it ends the invocation — Discord only allows follow-ups *after* that ACK, so ordered multi-message needs a deferred ACK + a second async invocation. That's not a one-liner, so per your call it's now tracked as its own issue: **#35** (deferred ACK → async self-invoke → `PATCH @original` + ordered follow-up `POST`s; no message bus). This PR keeps the explicit single-message truncation notice as the interim behaviour.

---

petbot/frontends/interactions/render.py
auzbuzzard: Follow-up on the chunking question: full ordered multi-message delivery (so a 10k-char result is sent in full, not truncated) is now tracked as **#35**. `chunk_text` in `core/text.py` is the shared primitive it'll use; the interactions adapter keeps the single-message truncation notice until #35 lands the deferred-ACK + async follow-up path.

---

Issue #38, opened 2026-06-12
Config &amp; startup hardening: pydantic-settings + per-frontend env validation


auzbuzzard:
Sub-issue of #28. Surfaced while planning the serverless deploy runbook, then broadened into a full audit of env-var injection + app startup.

## Problem (audit)
Configuration and startup were "all over the place":
1. **Frontend requirements conflated.** `Settings.from_env` *always* hard-required `DISCORD_TOKEN` (gateway-only), while `DISCORD_PUBLIC_KEY` (interactions-only) was checked ad-hoc deep inside `build_handler`. The Lambda would crash on boot or carry an **unused** bot token.
2. **Divergent startup paths.** Gateway (`__main__` → `load_dotenv` → `from_env` → `bootstrap.run` → `configure_logging`) vs. Lambda (`lambda_handler` → `from_env` → `_ensure_logging_configured` global). Logging-once logic and dotenv loading duplicated/divergent.
3. **Startup work on the request path.** `lambda_handler` re-parsed env + rebuilt the handler every invocation.
4. **Hand-rolled parsing** (`int()`, `LOG_FORMAT` case/membership, `or None`).
5. Stale `0004`→`0005` ADR reference in `pyproject.toml`.

## Fix (the correct, pythonic, platform-agnostic shape)
Adopt **pydantic-settings** (pydantic v2 is already a core dep): a shared `AppSettings` base + per-frontend subclasses that each validate exactly what they need, fail-fast.
- `GatewaySettings` → requires `DISCORD_TOKEN` (the parked /music worker).
- `InteractionsSettings` → requires **only** `DISCORD_PUBLIC_KEY`; the Lambda boots without the bot token.
- Settings + logging built **once per warm container** (cached cold-start), off the request path; drop the `_logging_configured` global and the ad-hoc public-key check.
- `.env` loading handled natively by pydantic-settings (no explicit `load_dotenv`).
- Config stays **agnostic to how the env was populated** (`op run`, plaintext `.env`, exported vars, Lambda/container injection).

## Done when
`InteractionsSettings` boots on `DISCORD_PUBLIC_KEY` alone (no `DISCORD_TOKEN`); the gateway still errors helpfully without its token; per-frontend unit tests cover both paths; all quality gates green. Prerequisite for a clean #30 deploy.

---
**Resolved by PR #40** (`Config & startup: pydantic-settings + per-frontend validation`).

Issue #35, opened 2026-06-10
Deferred-response mechanism + ordered multi-message delivery (interactions adapter)


auzbuzzard:
## Problem

The HTTP-Interactions / Lambda adapter (`petbot/frontends/interactions/`) can only deliver a **single message** in its immediate HTTP response, and that response *is* the Lambda's return value — returning it **ends the invocation**. Two limitations follow from the same gap:

1. **Long output is truncated.** Output longer than one Discord message (2000 chars) is capped with a truncation notice (`render.py::_truncate_to_single_message`). That's a stopgap — long output should be sent **in full, split across multiple messages, in order**.
2. **The 3s budget caps slow skills.** `handler.py` enforces `DEFAULT_SKILL_TIMEOUT_SECONDS` (2.5s) because Discord cancels any interaction not answered within 3s. Genuinely long-running skills (and boorus approaching the limit) need to run *after* the ACK.

Both are solved by the same missing piece: the **deferred-response mechanism**. This issue owns that mechanism *and* the ordered multi-message feature built directly on top of it.

> **Reference fix needed:** the code comments in `handler.py` / `render.py` currently point at `#33` for "the deferred follow-up path." That's a mis-reference — #33 is the music-remote-skill issue, a *consumer* of this mechanism, not its owner. Those comments should be updated to point here (tracked as a follow-up code change on the branch that implements this).

## Why it isn't a one-line change

For an **HTTP-interactions bot**, Discord only permits follow-up messages *after* the initial ACK — by which point the function has already exited. So deferred / ordered multi-message delivery cannot happen within a single Lambda invocation; it requires a second (asynchronous) execution context.

## Proposed design (no message bus)

1. **Webhook invocation** verifies the signature, then responds with a **deferred** ACK (`type 5`, `DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE`) as the HTTP body — within Discord's 3s deadline.
2. Before returning, it **async self-invokes** the same Lambda (`lambda:InvokeFunction`, `InvocationType=Event` — no SQS/EventBridge), passing `{application_id, interaction_token, command_data}`.
3. The **worker invocation** runs the skill (no longer bound by 3s), `chunk_text`s the output, then delivers it in order via Discord REST:
   - `PATCH /webhooks/{application_id}/{interaction_token}/messages/@original` with chunk 1 (fills the deferred placeholder),
   - `POST /webhooks/{application_id}/{interaction_token}` for chunks 2..N.

[...]

Issue #33, opened 2026-06-04
Future: re-activate music as a remote skill (cluster worker + message bus)


auzbuzzard:
Sub-issue of #28. **Deferred** — captured now so the seam is designed for, not bolted on. Re-enables `/music` once the stateless core is in steady state, with **zero changes** to the other skills.

## Shape
Music stays the one skill that genuinely needs a persistent **Gateway + voice (UDP)** connection, so it runs as its **own image on the homelab cluster**, and the Lambda core reaches it through a **message-bus queue** as a remote proxy — staying agnostic to where the skill lives.

- **`DispatchPort`** (new neutral port in `petbot/core/`, mirrors `VoicePort`): `async def dispatch(job) -&gt; None`. Interface only, no SDK.
- **`RemoteMusicSkill`** — implements the `Skill` interface; its `run()` validates args, **enqueues** `{guild, voice_channel, action, query, interaction_token/channel_id, …}`, and returns a **deferred ACK** ("🎵 on it"). Lives at the interactions composition root (infra), injected with the concrete queue client; core stays pure.
- **Music worker image** — the existing `petbot/frontends/discord/` gateway adapter trimmed to voice, packaged via the existing `Dockerfile` (FFmpeg + yt-dlp). Holds the gateway+voice connection, **consumes** the queue, runs the real (pure) `MusicSkill` over a real `VoicePort`, and posts results back to Discord via the interaction follow-up token or a bot-REST channel message.
- **Bus** — `aws_sqs_queue` (the stubbed block in #30). Worker **pulls** → no inbound to home, NAT-friendly, no tunnel.
- **Deploy** — GitOps: CI builds+pushes the image to GHCR and bumps the tag; **Argo CD/Flux in-cluster pulls and rolls** the `Deployment` (`replicas: 1` — gateway bots don't horizontally scale). Carries forward the gateway/FFmpeg ops from #17 (now scoped here).

## Dependency: deferred-response mechanism (#35)
`RemoteMusicSkill`'s **deferred ACK + interaction-token follow-up** is the generic deferred-response mechanism owned by **#35**, not something this issue defines. This issue is a *consumer* of that mechanism: it enqueues a job and the worker posts the eventual result back through the same `@original` PATCH / follow-up POST path. Build #35 first (or alongside); don't re-implement the ACK/follow-up plumbing here.

## Notes
[...]

Issue #17, opened 2026-05-30
Deployment &amp; secrets operations


auzbuzzard:
Part of #8 (item 6).

&gt; **Superseded / rescoped by the deployment epic #28.** The deployment approach pivoted from a single gateway container to a **serverless split**: the stateless core runs on **AWS Lambda** (infra-as-code in #30, secrets handled there), and the **gateway/FFmpeg** runtime now applies only to the **future music worker** on the cluster (#33). The items below are reorganized accordingly; this issue stays open as the home for **secrets/ops** that aren't owned by the Terraform stack.

## Now owned elsewhere
- Lambda infra + secret delivery (SSM/Secrets Manager, or 1Password→env at deploy) → **#30**
- Discord Developer Portal app, intents, command + endpoint registration → **#31**
- Gateway/FFmpeg `Dockerfile` runtime, GHCR image, GitOps deploy to the cluster → **#33**

## Still tracked here (cross-cutting secrets/ops)
- [ ] Create the 1Password vault items referenced by the `op://` paths in `.env.example` (used by local dev and the music worker; document the plaintext `.env` fallback for contributors without 1Password)
- [ ] Provision a **1Password Service Account** (`OP_SERVICE_ACCOUNT_TOKEN`) for headless secret resolution on any non-laptop host (CI, VPS, cluster) — the bootstrap secret that unlocks the `op://` references
- [ ] Decide secret delivery per runtime: Lambda (#30) vs. cluster music worker (#33, e.g. External Secrets / 1Password Operator vs. service-account token) — the app only reads `os.environ`, so this stays an ops choice with no code change
- [ ] Message Content intent note: only needed once the Phase B LLM layer (#15) lands

PR #41, opened 2026-06-13
Terraform: interactions Lambda + Function URL + IAM (#30)

### LINKED ISSUE #30: Terraform: AWS Lambda + Function URL + IAM (repo-owned deploy)

auzbuzzard:
## What & why

Repo-owned Terraform for the serverless HTTP-Interactions frontend (epic #28, **issue #30**): an AWS Lambda **container image** behind a **Function URL**, with least-privilege IAM. The Function URL output is what #31 sets as Discord's Interactions Endpoint URL. Decisions confirmed with @auzbuzzard: **container image**, **S3 + DynamoDB** state, **lint-only CI**.

I can't `apply` from CI's environment, but everything is **validated locally with terraform 1.9.8** (`fmt` clean, `validate` OK on both stacks).

## Layout (`deploy/terraform/`)

- **`main.tf`** — ECR repo, IAM role (CloudWatch Logs only), `aws_lambda_function` (arm64, `package_type=Image`, pinned to the image **digest**), `aws_lambda_function_url` (`AUTH_NONE`) + the required public-invoke permission.
- **`secrets.tf`** — reads **out-of-band SSM SecureString** params into the Lambda env: `DISCORD_PUBLIC_KEY` required, optional booru auth via an `ENV_NAME => ssm_name` map. No `DISCORD_TOKEN` (the Lambda never uses it, post-#38).
- **`versions.tf` / `providers.tf` / `variables.tf` / `outputs.tf`** — provider pin, S3 backend (configured at init via `backend.hcl`), inputs, and the `function_url` output.
- **`music.tf`** — commented future stubs: the #35 self-invoke permission and the #33 SQS music bus.
- **`bootstrap/`** — one-time stack (local state) that creates the encrypted+versioned state bucket and the lock table.
- **`../Dockerfile.lambda`** — Lambda image from `public.ecr.aws/lambda/python:3.12`, installs `.[interactions]`, CMD `petbot.frontends.interactions.app.lambda_handler`.
- **`README.md`** — full apply/teardown runbook (bootstrap → SSM params → two-step ECR/image/apply).
- **`.github/workflows/terraform.yml`** — `fmt -check` + `validate` (init `-backend=false`); no AWS creds needed.

## Flag for review (a correction to my earlier #30 note)

My planning comment on #30 said secrets would be kept **out of state**. In this implementation secrets are injected as **Lambda env vars from SSM data sources**, so the decrypted values *do* land in Terraform **state** — the standard pattern that keeps the app 100% `os.environ`/platform-agnostic. Mitigation: the bootstrapped state bucket is **encrypted + versioned + private**. Truly state-free secrets would need the function to **fetch SSM at runtime** (adds boto3 to the adapter); I scoped that out and noted it as a future hardening under secrets/ops #17. Happy to switch approaches if you'd rather keep state secret-free now.

Other notes:
- First apply is a documented **two-step** (create ECR → build/push image → full apply), because the function pins the image digest.
- CI is lint-only; a full **OIDC plan/apply** pipeline is a deliberate follow-up (needs an AWS OIDC role).

## Done when (per #30)
`terraform apply` produces a live Function URL serving the interactions handler — reproducible from the README; the apply itself runs from your AWS account.

Closes #30.

https://claude.ai/code/session_01FZR35H7aWgUw1Xquqc3W85

---
_Generated by [Claude Code](https://claude.ai/code/session_01FZR35H7aWgUw1Xquqc3W85)_

### DIFF
--- a/deploy/Dockerfile.lambda
+++ b/deploy/Dockerfile.lambda
@@ -9,7 +9,10 @@
 #
 #   docker build --platform linux/arm64 -f deploy/Dockerfile.lambda -t <ecr_url>:<tag> .
 #
-FROM public.ecr.aws/lambda/python:3.13
+# Match the runtime to what CI tests and the toolchain targets (pyproject:
+# requires-python, mypy, ruff are all 3.12). Shipping a newer interpreter than
+# CI exercises would deploy an untested runtime — keep them in lockstep.
+FROM public.ecr.aws/lambda/python:3.12
 
 # ${LAMBDA_TASK_ROOT} is /var/task, already the WORKDIR and on sys.path.
 COPY pyproject.toml README.md ./
--- a/deploy/Dockerfile.lambda
+++ b/deploy/Dockerfile.lambda
@@ -9,7 +9,7 @@
 #
 #   docker build --platform linux/arm64 -f deploy/Dockerfile.lambda -t <ecr_url>:<tag> .
 #
-FROM public.ecr.aws/lambda/python:3.12
+FROM public.ecr.aws/lambda/python:3.13
 
 # ${LAMBDA_TASK_ROOT} is /var/task, already the WORKDIR and on sys.path.
 COPY pyproject.toml README.md ./
--- a/deploy/Dockerfile.lambda
+++ b/deploy/Dockerfile.lambda
@@ -1,0 +1,24 @@
+# PetBot HTTP-Interactions Lambda image.
+#
+# A container image (not a zip) because the adapter pulls in C-extensions
+# (pynacl for Ed25519, numexpr/numpy for /math); the AWS Lambda Python base
+# image gives a matching manylinux runtime with no wheel wrangling.
+#
+# Build from the repository root (the build context must see pyproject.toml and
+# petbot/), targeting arm64 to match the Lambda's Graviton architecture:
+#
+#   docker build --platform linux/arm64 -f deploy/Dockerfile.lambda -t <ecr_url>:<tag> .
+#
+FROM public.ecr.aws/lambda/python:3.12
+
+# ${LAMBDA_TASK_ROOT} is /var/task, already the WORKDIR and on sys.path.
+COPY pyproject.toml README.md ./
+COPY petbot ./petbot
+
+# `[interactions]` pulls in pynacl; the base dependencies bring pydantic,
+# pydantic-settings, httpx, numexpr, etc. No secrets are baked in — they are
+# supplied as environment variables by Terraform (see deploy/terraform).
+RUN pip install --no-cache-dir ".[interactions]"
+
+# module.function the Lambda runtime invokes per request.
+CMD ["petbot.frontends.interactions.app.lambda_handler"]
--- a/deploy/terraform/main.tf
+++ b/deploy/terraform/main.tf
@@ -1,5 +1,3 @@
-data "aws_caller_identity" "current" {}
-
 # --- Container image registry -------------------------------------------------
 
 resource "aws_ecr_repository" "this" {
@@ -60,6 +58,10 @@ resource "aws_lambda_function" "this" {
   environment {
     variables = local.lambda_env
   }
+
+  # Ensure our retention-managed log group exists before the function can be
+  # invoked, so Lambda doesn't auto-create an unmanaged, never-expiring one.
+  depends_on = [aws_cloudwatch_log_group.lambda]
 }
 
 resource "aws_lambda_function_url" "this" {
--- a/deploy/terraform/main.tf
+++ b/deploy/terraform/main.tf
@@ -1,0 +1,78 @@
+data "aws_caller_identity" "current" {}
+
+# --- Container image registry -------------------------------------------------
+
+resource "aws_ecr_repository" "this" {
+  name                 = var.name_prefix
+  image_tag_mutability = "MUTABLE"
+  force_delete         = true
+
+  image_scanning_configuration {
+    scan_on_push = true
+  }
+}
+
+# Resolve the tag to an immutable digest so re-pushing the same tag and
+# re-applying actually updates the function. Requires the image to exist before
+# the main apply (see the two-step bootstrap in README).
+data "aws_ecr_image" "this" {
+  repository_name = aws_ecr_repository.this.name
+  image_tag       = var.image_tag
+}
+
+# --- Execution role -----------------------------------------------------------
+
+data "aws_iam_policy_document" "assume" {
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role" "lambda" {
+  name               = "${var.name_prefix}-role"
+  assume_role_policy = data.aws_iam_policy_document.assume.json
+}
+
+# CloudWatch Logs only — secrets are injected as env at deploy, so the function
+# needs no runtime SSM/KMS access. (A future runtime-fetch model would add a
+# scoped ssm:GetParameter policy here.)
+resource "aws_iam_role_policy_attachment" "logs" {
+  role       = aws_iam_role.lambda.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+}
+
+# --- Function + public URL ----------------------------------------------------
+
+resource "aws_lambda_function" "this" {
+  function_name = var.name_prefix
+  role          = aws_iam_role.lambda.arn
+  package_type  = "Image"
+  image_uri     = "${aws_ecr_repository.this.repository_url}@${data.aws_ecr_image.this.image_digest}"
+  architectures = ["arm64"]
+  memory_size   = var.memory_size
+  timeout       = var.timeout
+
+  environment {
+    variables = local.lambda_env
+  }
+}
+
+resource "aws_lambda_function_url" "this" {
+  function_name      = aws_lambda_function.this.function_name
+  authorization_type = "NONE"
+}
+
+# AUTH_NONE still requires an explicit public-invoke permission; the security
+# boundary is the Ed25519 signature the handler verifies, not IAM.
+resource "aws_lambda_permission" "function_url" {
+  statement_id           = "FunctionURLAllowPublicAccess"
+  action                 = "lambda:InvokeFunctionUrl"
+  function_name          = aws_lambda_function.this.function_name
+  principal              = "*"
+  function_url_auth_type = "NONE"
+}
--- a/deploy/terraform/secrets.tf
+++ b/deploy/terraform/secrets.tf
@@ -1,0 +1,37 @@
+# Runtime secrets are stored as SSM SecureString parameters created out-of-band
+# (see README) and read here to populate the Lambda's environment. The bot only
+# ever reads os.environ, so the app stays platform-agnostic.
+#
+# Trade-off (flagged for review): injecting via the Lambda environment means the
+# decrypted values land in the Lambda's config and therefore in Terraform state.
+# The S3 backend is encrypted, versioned, and access-controlled to contain this.
+# Keeping state fully secret-free would require the function to fetch SSM at
+# runtime instead — a documented future hardening (tracked with secrets/ops #17).
+
+data "aws_ssm_parameter" "public_key" {
+  name            = var.public_key_ssm_parameter
+  with_decryption = true
+}
+
+data "aws_ssm_parameter" "booru" {
+  for_each        = var.booru_ssm_parameters
+  name            = each.value
+  with_decryption = true
+}
+
+locals {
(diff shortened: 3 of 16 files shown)


deploy/terraform/variables.tf
```diff
@@ -0,0 +1,71 @@
+variable "aws_region" {
+  type        = string
+  description = "AWS region for all resources."
+  default     = "us-east-1"
```
auzbuzzard, line 13: why us-east-1 as default? 

Also educate me who get the ultimate decision on where the lambda is deployed? I assume all resources has to be in the same region.

I would prefer to have resources in the US and also in the cheapest region. Do research best region.

deploy/terraform/secrets.tf
```diff
@@ -0,0 +1,37 @@
+# Runtime secrets are stored as SSM SecureString parameters created out-of-band
+# (see README) and read here to populate the Lambda's environment. The bot only
+# ever reads os.environ, so the app stays platform-agnostic.
+#
+# Trade-off (flagged for review): injecting via the Lambda environment means the
+# decrypted values land in the Lambda's config and therefore in Terraform state.
+# The S3 backend is encrypted, versioned, and access-controlled to contain this.
+# Keeping state fully secret-free would require the function to fetch SSM at
+# runtime instead — a documented future hardening (tracked with secrets/ops #17).
```
auzbuzzard, line 9: explain the tradeoff. Are you saying we are passing secrets to the lambda function as env var, thus it's exposed in the environment? Is this advisable? What's the correct end state for this?

deploy/terraform/main.tf
```diff
@@ -0,0 +1,78 @@
+data "aws_caller_identity" "current" {}
```
auzbuzzard: Is this needed?

deploy/Dockerfile.lambda
```diff
@@ -0,0 +1,24 @@
...
+FROM public.ecr.aws/lambda/python:3.13
+
+# ${LAMBDA_TASK_ROOT} is /var/task, already the WORKDIR and on sys.path.
+COPY pyproject.toml README.md ./
+COPY petbot ./petbot
+
+# `[interactions]` pulls in pynacl; the base dependencies bring pydantic,
+# pydantic-settings, httpx, numexpr, etc. No secrets are baked in — they are
+# supplied as environment variables by Terraform (see deploy/terraform).
+RUN pip install --no-cache-dir ".[interactions]"
```
auzbuzzard, line 24: Is pip better or worse than uv in a lambda environment?

deploy/terraform/terraform.tfvars.example
```diff
@@ -0,0 +1,19 @@
+# Copy to terraform.tfvars and adjust. No secrets live here — secret *values*
+# are stored as SSM SecureString parameters (see README), referenced by name.
+
+aws_region = "us-east-1"
```
auzbuzzard, line 6: Same as the other comment. What aws region should be default?

deploy/Dockerfile.lambda
```diff
@@ -0,0 +1,24 @@
...
+# A container image (not a zip) because the adapter pulls in C-extensions
+# (pynacl for Ed25519, numexpr/numpy for /math); the AWS Lambda Python base
+# image gives a matching manylinux runtime with no wheel wrangling.
+#
+# Build from the repository root (the build context must see pyproject.toml and
+# petbot/), targeting arm64 to match the Lambda's Graviton architecture:
+#
+#   docker build --platform linux/arm64 -f deploy/Dockerfile.lambda -t <ecr_url>:<tag> .
+#
+FROM public.ecr.aws/lambda/python:3.13
```
auzbuzzard: Why 13? Isn't the repo locked to 3.12?

auzbuzzard (commented):
1. does terraform handles resource cleanup (e.g. ecr retention policy, cloudwatch retention policy? What about cost alerts/limits?
2.

auzbuzzard:
Thanks for the thorough review — addressed in `30e2b18`. Point-by-point:

**Retention & cost cleanup** (general review) — **was missing; now added.**
- `retention.tf`: an **ECR lifecycle policy** (expire untagged after `ecr_untagged_expire_days`, keep last `ecr_keep_last_images`) and an explicit **`aws_cloudwatch_log_group`** with `log_retention_days`. The latter matters twice: logs no longer accumulate forever, and `terraform destroy` now *owns* the group (Lambda's auto-created one never expires and TF wouldn't manage it). The function `depends_on` it so Lambda can't race to create an unmanaged one.
- `budget.tf`: an **opt-in** monthly AWS Budgets alert (set `monthly_budget_usd` + `budget_alert_emails` to enable; off by default). Note AWS has **no hard "stop spending" cap** — Budgets is an alert, not a throttle. At our scale spend is ~$0.
- (Your point **2.** came through empty — resend if there was more.)

**Region — `variables.tf` & `terraform.tfvars.example`.** Researched it: **us-east-1 is AWS's cheapest US region** (the pricing baseline; us-east-2 / us-west-2 match it, us-west-1 is pricier). **You own the decision** via the single `aws_region` variable — every regional resource here (Lambda, ECR, CloudWatch, SSM, state bucket) inherits it; IAM is global. Kept us-east-1 as default and documented *why* on the variable: besides being cheapest, it's where **ACM/CloudFront** for the future custom domain (#31/#32) must live, so it avoids cross-region friction. Caveat now noted: the SSM params must be created in that region, and `backend.hcl`'s region should match.

**`Dockerfile.lambda` — why 3.13 if the repo is 3.12?** Good catch. `requires-python = ">=3.12"` is a *floor*, not a lock, so 3.13 is technically allowed — **but** CI, mypy, and ruff all target 3.12, so shipping 3.13 would run prod on an interpreter we never test. **Reverted the image to `python:3.12`** for test==prod parity. If you'd rather be on 3.13, that's a separate repo-wide bump (image + CI + mypy/ruff targets + `requires-python`) I'm happy to do as its own PR.

**`Dockerfile.lambda` — pip vs uv?** At **runtime, no difference** — both install identical wheels; the Lambda executes the same bytes, same cold-start. `uv` would only speed up `docker build`. Not worth adding a tool to the image for this bot's occasional builds, but I'll switch if you want faster image builds in the CI deploy pipeline (#42). No change made.

**`secrets.tf` — the tradeoff.** Yes: secrets are injected as Lambda **environment variables**, so they're visible in the function's configuration (`get-function-configuration`, console) and embedded in Terraform **state**. Advisable? **For this app, it's mild** — `DISCORD_PUBLIC_KEY` is literally public (it only verifies inbound signatures), and the booru keys are optional/low-sensitivity. Lambda also encrypts env vars at rest with KMS. The **correct end state** for genuinely sensitive secrets is the function **fetching from SSM/Secrets Manager at runtime** (or the AWS Parameters & Secrets Lambda extension), so plaintext never lands in env/config/state — that adds `boto3` to the adapter and is tracked as hardening under #17. So: acceptable now given *what* these secrets are; runtime-fetch is the documented upgrade.

**`main.tf:1` — is `aws_caller_identity` needed?** No — unused dead code (leftover from a scoped-policy idea). **Removed.**

All re-validated locally (terraform 1.13.1 / aws v6.50.0): `fmt` clean, `validate` OK.

---

deploy/terraform/variables.tf
auzbuzzard: Researched it (fixed in `30e2b18`): **us-east-1 is AWS's cheapest US region** (the pricing baseline; us-east-2 / us-west-2 match it, us-west-1 is pricier).

**Who decides:** you do — via the single `aws_region` variable. Every *regional* resource here (Lambda, ECR, CloudWatch, SSM, the state bucket) inherits it, so yes they're all in one region; IAM is global and unaffected. Kept us-east-1 as default and documented *why* on the variable: besides being cheapest, it's where **ACM/CloudFront** for the future custom domain (#31/#32) must live, avoiding cross-region friction. Caveats now noted: the SSM params must be created in this region, and `backend.hcl`'s region should match.

---

deploy/terraform/secrets.tf
auzbuzzard: Yes — secrets are injected as Lambda **environment variables**, so they're visible in the function's configuration (`aws lambda get-function-configuration`, the console) and embedded in Terraform **state**.

Advisable? **For this app it's mild:** `DISCORD_PUBLIC_KEY` is literally public (it only verifies inbound signatures — leaking it is harmless), and the booru keys are optional/low-sensitivity. Lambda also encrypts env vars at rest with KMS.

**Correct end state** for genuinely sensitive secrets: have the function **fetch from SSM/Secrets Manager at runtime** (or use the AWS Parameters & Secrets Lambda extension), so plaintext never lands in env/config/state. That adds `boto3` to the adapter and is tracked as hardening under #17. So: acceptable now given *what* these secrets are; runtime-fetch is the upgrade path.

---

deploy/terraform/main.tf
auzbuzzard: No — it was unused dead code (a leftover from a scoped-policy idea). **Removed** in `30e2b18`.

---

deploy/Dockerfile.lambda
auzbuzzard: At **runtime, no difference** — pip and uv install the same wheels, so the Lambda executes identical bytes with the same cold-start. `uv` would only make `docker build` faster. Not worth adding a tool to the image for this bot's occasional builds, but I'm happy to switch to `uv` if you want quicker image builds in the #42 CI pipeline. Left as `pip` for now.

---

deploy/Dockerfile.lambda
auzbuzzard: Good catch. `requires-python = ">=3.12"` is a *floor*, not a lock, so 3.13 is technically allowed — **but** CI, mypy, and ruff all target 3.12, so shipping 3.13 would run prod on an interpreter we never test. **Reverted the image to `python:3.12`** in `30e2b18` for test==prod parity. If you'd rather be on 3.13, that's a separate repo-wide bump (image + CI + mypy/ruff targets + `requires-python`) I'll do as its own PR.

---

deploy/terraform/terraform.tfvars.example
```diff
@@ -0,0 +1,30 @@
...
+# }
+
+# --- Retention / cost (sensible defaults; override to taste) ---
+# log_retention_days       = 30
+# ecr_keep_last_images     = 10
+# ecr_untagged_expire_days = 7
+
+# Optional monthly cost alert (AWS Budgets). Both must be set to enable it.
+# monthly_budget_usd  = 5
+# budget_alert_emails = ["you@example.com"]
```
auzbuzzard, line 30: so where is the actual email defined?

auzbuzzard: It isn't defined anywhere in the repo — **you** define it, in your own `terraform.tfvars` (the local copy you make of this `.example`; it's git-ignored). You'd put:

```hcl
budget_alert_emails = ["you@yourdomain.com"]
monthly_budget_usd  = 5
```

That binds to the `budget_alert_emails` **variable** (`variables.tf`), which `budget.tf` passes to the budget's `subscriber_email_addresses`. Equivalent alternatives if you don't want it in a file: `terraform apply -var 'budget_alert_emails=["you@…"]'`, or an env var `export TF_VAR_budget_alert_emails='["you@…"]'`.

This `.example` only carries the `you@example.com` placeholder so nothing real is ever committed. And it's entirely optional — leave `monthly_budget_usd` at 0 and no budget (and no email) is used at all.

---

Issue #15, opened 2026-05-30
Phase B — LLM agent layer (provider-agnostic)


auzbuzzard:
**Epic** — part of #8 (item 4). See `docs/adr/0002-deferred-llm.md`.

Add an LLM "agent" layer so a user can chat with the bot and it can call skills through a tool-calling loop. It lives in `petbot/core/llm/` (platform-neutral, like skills) and reuses the existing neutral `Skill` / `SkillRegistry` / `SkillResult` contract — **no changes to skills required**.

**Provider choice is pending (cost-driven)** — the design keeps the provider a config option, not a hard dependency.

### 1. Provider-agnostic LLM client interface
Define a minimal `Protocol` for an LLM client — a chat/completion call that supports **tool/function calling** — with an adapter per provider (OpenAI-compatible, Gemini, Groq/OpenRouter, local via Ollama, …) selected by config. No hard dependency on any single SDK; no provider chosen yet (cost-driven). Lives in `petbot/core/llm/`, platform-neutral.

### 2. Tool-calling loop over the `SkillRegistry`
Format each registered skill's `input_schema` (already JSON Schema) into the chosen provider's tool format, dispatch tool calls to `skill.run(args, ctx)`, feed results back, and return the final `SkillResult`. Reuses the same neutral `SkillRegistry` entrypoint every frontend uses — so the LLM gets the exact set of skills a given frontend exposes (via `available_for(capabilities)`).

### 3. Conversation sessions + Discord chat entrypoint
Persist per-conversation history keyed by the neutral `conversation_id` already on `SkillContext` (so chat works identically across frontends). Add the Discord conversational entrypoint — @mention / thread reply — in the Discord adapter, and enable the **Message Content** privileged intent (only required at this point; slash commands never needed it). Decide the session store (in-memory to start; pluggable later).

## Done when
A user can @mention/thread the bot on Discord, it answers conversationally, and it can invoke skills (e.g. "find me a pony pic") via the LLM, with the provider selectable by config.

PR #53, opened 2026-06-17
PetBot 2.1: A3a — typed edge→worker dispatch + LLM chat agent (#15)

auzbuzzard:
## What

Lands the LLM/agent layer (#15) by reimagining PetBot as **A3a**: a thin, always-on Discord **edge** that holds the gateway and runs no skills, dispatching to skill **workers** that do. One uv workspace, many independently-installable packages under the `petbot.*` namespace (PEP 420), wired by a **typed calling pattern** that `mypy --strict` checks across package boundaries.

Replaces the earlier groundwork branch (#52), which sprawled. This branch is a clean rebuild from the last good tooling commit.

## The calling pattern

```python
# the edge holds a typed client; the wire + skill-name string are encapsulated
skills: Skills = RemoteSkills(HttpTransport(url, client))
await skills.chat(ChatArgs(message=text), ctx)   # fully typed across packages
```

The client serialises a `SkillCall`; the worker re-validates the args against the skill's `args_model` and runs it. No hand-rolled wire layer — pydantic models serialise themselves.

## Packages

| Package | Role |
|---|---|
| `petbot.domain` | Frozen pydantic kernel: `Skill[ArgsT]`, `SkillResult`, `SkillContext`, `SkillCall`, `Transport`, ports. Depends on nothing first-party. |
| `petbot.types` | The typed surface the edge imports without skills: `*Args` models + the `Skills` Protocol. |
| `petbot.platform` | `Worker`, `RemoteSkills`/`LocalSkills`, `HttpTransport`/`LambdaTransport`. |
| `petbot.skills.{math,booru,music,chat}` | One skill each. booru ports the legacy engine verbatim; **chat is a pydantic-ai agent whose tools are its sibling skills**. |
| `petbot.discord` | The edge: `@mention` → `skills.chat(...)` → render. |
| `petbot.workers.{brain,music}` | Deploy bundles (brain = math+booru+chat; music = gateway + voice host). |

## LLM (issue #15)

- **chat** is a normal skill built on **pydantic-ai**. Each sibling skill is an `@agent.tool` whose argument type *is* the `petbot.types.*Args` model — one declaration feeds the typed client, the worker's validation, and the LLM tool schema.
- **Provider-agnostic, no vendor lock-in** (`CHAT_PROVIDER`): Amazon Bedrock for prod, OpenRouter (free models) for dev. Model ids are config, never code. (Note: Bedrock does not host Gemma — Gemma-free is available via OpenRouter; ADR 0007.)
- Tests drive the agent with pydantic-ai's `TestModel`/`FunctionModel`, so CI exercises the tool-calling path with **no live LLM and no secrets**.

## Verification (all green, offline)

`ruff` · `mypy --strict` (62 files) · `import-linter` (4 boundary contracts) · **53 tests**. `pyrightconfig.json` fixes Pylance for the editable namespace installs; `uv.lock` regenerated (`--locked` passes).

## Docs

Bump to 2.1. ADR 0007 (LLM agent — pydantic-ai) added; ADR 0002 (deferred LLM) superseded. README / AGENTS / architecture / contributing updated for the A3a layout.

## Known follow-up (not in this PR)

`deploy.yml` + `terraform/` still target the deleted serverless *interactions* Lambda. The deploy pipeline needs rewriting for the A3a topology (brain-worker Lambda + small always-on edge VM) — to be done after review. It does not affect CI (only triggers on release).

Closes the work tracked by #52. Relates to #15.

---
_Generated by [Claude Code](https://claude.ai/code/session_018AE3E1MRysoTBHxgQhH1rA)_


platform/petbot/platform/skills.py
```diff
@@ -0,0 +1,71 @@
...
+        return await self._call("e621", args, ctx)
+
+    async def music(self, args: MusicArgs, ctx: SkillContext) -> SkillResult:
+        return await self._call("music", args, ctx)
+
+    async def chat(self, args: ChatArgs, ctx: SkillContext) -> SkillResult:
+        return await self._call("chat", args, ctx)
+
+
+class LocalSkills:
```
auzbuzzard: Why do we need LocalSkills? Whos calling it? Can the impl not be combined with RemoteSkills? It's very similar.

skills/booru/petbot/skills/booru/settings.py
```diff
@@ -0,0 +1,36 @@
...
+unauthenticated, lower-rate-limit requests.
+"""
+
+from __future__ import annotations
+
+from pydantic import field_validator
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+#: Descriptive User-Agent sent to booru APIs; a spoofed one gets blocked.
+DEFAULT_USER_AGENT = "PetBot/2.1 (https://github.com/auzbuzzard/petbot)"
```
auzbuzzard: The PetBot version string can it be built dynamically from the project config?

src/petbot/skills/chat/settings.py
```diff
@@ -0,0 +1,39 @@
...
+        frozen=True,
+    )
+
+    provider: ChatProvider = "bedrock"
+    #: Bedrock model id (used when ``provider == "bedrock"``).
+    bedrock_model: str = "amazon.nova-lite-v1:0"
+    #: OpenRouter model id (used when ``provider == "openrouter"``).
+    openrouter_model: str = "google/gemma-2-9b-it:free"
+    #: OpenRouter API key; required only when ``provider == "openrouter"``.
+    openrouter_api_key: str | None = None
```
auzbuzzard: Review if these are the latest model. I remember openrouter has gemma3 for free probably.
Also, why are we hardcoding model name in code??? Best should these be default values only?

auzbuzzard (commented):
1. what is workers/brain? Can we rename brain to "core" or something more meaningful + descriptive of the actual membership of that worker?

2. Stop leaking conversations and replies into the docs and code comments. Don't use jargons like A3a, or anything that is time-variant. Review all docs and comments.

3. Why do we need to nest the dir like this? e.g. frontent/discord/petbot/discord . What's going on here???

4. Review all namings and verbiages and jargons. e.g. are we using frontend or edge? EdgeSettings vs frontend/ impl. 

5. Review all loggings and loggers. Are they still following the logging system I defined in #25? I see the logging config is deleted and not all packages use the right logging code style.

src/petbot/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""PetBot: a Discord bot — an edge that dispatches to skill workers."""
+
+from __future__ import annotations
+
+__version__ = "2.1.0"
```
auzbuzzard: Why are we redefining version? Shouldn't there be only 1 authoritative version number ?

src/petbot/edge/settings.py
```diff
@@ -0,0 +1,40 @@
...
+        env_file=".env", env_file_encoding="utf-8", extra="ignore", frozen=True
+    )
+
+    discord_token: str
+    dev_guild_id: int | None = None
+    log_level: str = "INFO"
+
+    transport: TransportKind = "http"
+    #: Worker HTTP endpoint (used when ``transport == "http"``).
+    worker_url: str = "http://localhost:8000/dispatch"
```
auzbuzzard: Why do we have url hard coded? And where is local host coming from?

src/petbot/skills/chat/agent.py
```diff
@@ -0,0 +1,81 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+from pydantic_ai import Agent, RunContext
+
+from petbot.domain import SkillContext, SkillResult
+from petbot.types import BooruArgs, MathArgs, Skills
+
+SYSTEM_PROMPT = (
```
auzbuzzard: Why is system prompt hard coded? Should it be a file or an env? What's the correct thing to do here?

src/petbot/workers/music/settings.py
```diff
@@ -0,0 +1,22 @@
...
+    model_config = SettingsConfigDict(
+        env_prefix="music_",
+        env_file=".env",
+        env_file_encoding="utf-8",
+        extra="ignore",
+        frozen=True,
+    )
+
+    discord_token: str
+    dispatch_host: str = "0.0.0.0"
```
auzbuzzard: Same, why hard coded values?

platform/petbot/platform/skills.py
auzbuzzard: Done in `674d132`. Collapsed into one `SkillsClient` over a `Transport`; the in-process case is now a `LocalTransport` (calls `Worker.run` directly, **zero** serialization), and `SkillCall` carries the typed args instead of pre-serialized JSON. `LocalSkills`/`RemoteSkills` are deleted.

---

skills/booru/petbot/skills/booru/settings.py
auzbuzzard: Done. `DEFAULT_USER_AGENT` now builds from `importlib.metadata.version("petbot")` (`674d132`) — no literal. And the package's own `__version__` is single-sourced from `pyproject.toml` the same way in `189e07a`.

---

src/petbot/skills/chat/settings.py
auzbuzzard: These are pydantic-settings **defaults**, not values fixed in code — overridable via `CHAT_OPENROUTER_MODEL` / `CHAT_BEDROCK_MODEL`. Refreshed the dev default to the free `google/gemma-3-27b-it:free` (`674d132`). (Note: Bedrock doesn't host Gemma — that's why the Bedrock default is Amazon Nova Lite and Gemma lives on the OpenRouter side.)

---

src/petbot/__init__.py
auzbuzzard: Fixed in `189e07a`. `__version__ = importlib.metadata.version("petbot")` — `pyproject.toml` is now the single authoritative version; the duplicate literal is gone.

---

src/petbot/edge/settings.py
auzbuzzard: It's an overridable default (`WORKER_URL`), not a constant — and only consulted when `transport == "http"`; production uses `transport=lambda`, which never reads it. The `localhost` came from the core dev server's bind. Aligned it to that exactly — `http://127.0.0.1:8000/dispatch` (the worker binds `127.0.0.1:8000`, see `WORKER_HOST`/`WORKER_PORT` in `workers/core/__main__.py`) — and documented it in `189e07a`.

---

src/petbot/skills/chat/agent.py
auzbuzzard: Good call — it's product config, not logic. Moved it to a setting in `189e07a`: `ChatSettings.system_prompt`, overridable via `CHAT_SYSTEM_PROMPT`; `build_agent(system_prompt)` now takes it as a parameter and the default text lives in `ChatSettings`. An env var handles even long prompts; for a very large prompt you'd read a file's contents into it at deploy time. Documented in `.env.example`.

---

src/petbot/workers/music/settings.py
auzbuzzard: Same as the edge — these are overridable defaults (`MUSIC_DISPATCH_HOST` / `MUSIC_DISPATCH_PORT`), now documented in `189e07a`, including why the bind defaults to `0.0.0.0` (a network service the edge reaches across the network). If you'd rather network addresses be **required** (no default → fail-fast if unset, so a prod process can't silently fall back), say the word and I'll drop the defaults on these two fields.

---

auzbuzzard: yeah drop all defaults, including the 127.0.0.1 . Fail fast, app runner should provide the values instead of having dangling network hosts scattered everywhere.

src/petbot/skills/chat/settings.py
auzbuzzard: Do we need to write defaults in the code? Don't the app runner provide the value? Is there a reason we put it here? To me it smells having model names anywhere in the actual code.

src/petbot/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from __future__ import annotations
+
+from importlib.metadata import PackageNotFoundError, version
+
+try:
+    #: Single authoritative version — declared once in ``pyproject.toml`` and read
+    #: back from the installed package metadata, never duplicated as a literal.
+    __version__ = version("petbot")
+except PackageNotFoundError:  # pragma: no cover - only when run from a non-install
+    __version__ = "0+unknown"
```
auzbuzzard: Why do we need a __version__ here at all? Like, why? don't pyproject manage this? What's the point of having another variable mirroring it at random???

src/petbot/skills/chat/settings.py
auzbuzzard: Agreed — dropped the model defaults entirely in `5399b3f`. `bedrock_model` / `openrouter_model` are now `str | None = None` (no model name anywhere in code); the runner supplies `CHAT_BEDROCK_MODEL` / `CHAT_OPENROUTER_MODEL`, and `build_model` fails fast with a clear error if the selected provider's model id is unset.

---

src/petbot/workers/music/settings.py
auzbuzzard: Done in `5399b3f`. `dispatch_host` / `dispatch_port` are now required (no default) — the bind is always an explicit choice from the runner. Applied the same rule to the edge's `worker_url` (no default; `setup_hook` raises if `transport=http` and it's unset). No dangling hosts left in code.

---

src/petbot/__init__.py
auzbuzzard: You're right — removed `__version__` entirely in `5399b3f`. `pyproject.toml` is the sole version; nothing internal referenced `petbot.__version__` (the booru User-Agent queries `importlib.metadata.version("petbot")` directly), so there's no mirror at all now.

---

src/petbot/skills/chat/settings.py
auzbuzzard: Is None the correct way of doing this? Code would want a set variable, what's the correct pythonic way of handling env variables such as this? Review the other str | None type configs as well.

auzbuzzard: Addressed in `99ddb85`. Short answer: `None` is correct only for *genuinely optional* values; for *required* config it's a smell that pushes guards to the call site.

Two categories, applied across the settings:

- **Genuinely optional → kept `X | None = None`:** booru credentials, `dev_guild_id`, `Post.total`. Here `None` honestly means "absent / unauthenticated".
- **Required → no default, no `None` (fail fast at construction):**
  - chat: collapsed `bedrock_model` + `openrouter_model` into one **required** `model: str` (`CHAT_MODEL`) — exactly one model is ever active, so there's no `None` and no model name in code.
  - For values required only *in a mode* (the OpenRouter key; the transport target), a pydantic `@model_validator(mode="after")` enforces the rule at load:
    ```python
    @model_validator(mode="after")
    def _require_openrouter_key(self) -> Self:
        if self.provider == "openrouter" and not self.openrouter_api_key:
            raise ValueError("CHAT_PROVIDER=openrouter requires CHAT_OPENROUTER_API_KEY.")
        return self
    ```
    The failure now happens at settings load, not deep in `build_model`/`setup_hook`, which just trust the validated object.

`worker_url`/`worker_lambda` stay typed `str | None` (each is genuinely absent in the other transport mode), but the edge validator guarantees the *active* one is set and the call site narrows with `assert`.

One judgment call I made: I merged `CHAT_BEDROCK_MODEL`/`CHAT_OPENROUTER_MODEL` into a single `CHAT_MODEL`. If you'd rather keep two explicitly-named env vars (with the same validator), say so and I'll split them back.

---

src/petbot/edge/settings.py
```diff
@@ -0,0 +1,50 @@
...
+
+    discord_token: str
+    dev_guild_id: int | None = None
+    log_level: str = "INFO"
+
+    transport: TransportKind = "http"
+    #: Worker HTTP endpoint — present iff ``transport == "http"`` (``WORKER_URL``).
+    worker_url: str | None = None
+    #: Worker Lambda function name — present iff ``transport == "lambda"`` (``WORKER_LAMBDA``).
+    worker_lambda: str | None = None
```
auzbuzzard: More of these str | None. This is code smell. Review all of these patterns and see if a bag of values should be done correctly, so only the right properties are exposed.

auzbuzzard: Done in `092df3a` — modelled both config bags as discriminated (sum-type) unions, so only the active variant's fields exist (no `str | None`, no cross-mode fields):

- **`EdgeSettings.worker`** = `HttpWorker{url}` | `LambdaWorker{function_name}`, tagged by `kind`. `setup_hook` `match`es it (no `assert`/`None`). Env is nested: `WORKER__KIND=http` + `WORKER__URL=…`, or `WORKER__KIND=lambda` + `WORKER__FUNCTION_NAME=…`.
- **`ChatSettings.llm`** = `BedrockModel{model}` | `OpenRouterModel{model, api_key}`, tagged by `kind`. `build_model` `match`es it; the OpenRouter API key is a *required field of that variant only*. Env: `CHAT_LLM__KIND` + `CHAT_LLM__MODEL` (+ `__API_KEY`).

The `model_validator`s and the conditional `str | None` fields are gone. I left `| None` only where a value is *genuinely* optional (booru creds, `dev_guild_id`). `.env.example` updated to the nested vars.

---

PR #54, opened 2026-06-19
feat(deploy): retarget AWS stack to the 2.1 edge + core-worker topology

auzbuzzard:
Stacked on #53 — **base is `claude/petbot-a3a-2.1`**, so merging this lands the deploy rewrite *into* the #53 branch. Master only ever flips interactions→fully-migrated in one shot (merging #53 alone would auto-fire the stale `Deploy` on a half-migrated tree).

## Why
The `deploy/` stack still described the **deleted** serverless *interactions* design: a public Function-URL Lambda taking Discord webhooks, Ed25519 signature verification, and slash-command registration. #53 replaced that with **edge → worker**, so the deploy must follow: a **private core-worker Lambda** invoked by an **always-on edge**.

## What

**App (1 line + test)**
- `workers/core/handler.py` returns the **bare `SkillResult`** instead of an API-Gateway envelope, so the edge's `LambdaTransport` parses the boto3-invoke `Payload` directly.

**Core worker (Lambda)**
- Removed the Function URL, public invoke permission, and `DISCORD_PUBLIC_KEY` — the worker is **private**, invoked by the edge via **boto3 + IAM** (no public, unauthenticated LLM-cost surface).
- Chat-LLM config injected as nested env (`CHAT_LLM__KIND/MODEL[/API_KEY]`); **bedrock** gets a scoped `bedrock:InvokeModel` exec-role policy, **openrouter** reads its key from SSM. `timeout` 10→60.

**Edge (always-on gateway holder)**
- New **AWS Lightsail Container Service** (`nano`, amd64), env-configured, **no public endpoint** (outbound-only). Edge baseline measured ≈ 76 MB RSS; `micro` is the one-line fallback.
- A **scoped `petbot-edge` IAM user** whose only permission is `lambda:InvokeFunction` on the core worker; its access key is injected into the container env.
- **Image delivery = ECR-pull:** both images live in **ECR**; the Lightsail service pulls the edge image via private-registry access (an image-puller principal granted read on the edge repo in `edge.tf`). No `lightsailctl`, no image-ref scraping. *(This is the [official, IaC-supported path](https://docs.aws.amazon.com/lightsail/latest/userguide/amazon-lightsail-container-service-ecr-private-repo-access.html); the lightsail-registry push is positioned for local-machine pushes.)*

**Stack / CI**
- One `name_prefix` derives `-core` / `-edge`; state key → `petbot/terraform.tfstate`; SSM → `/petbot/{core,edge}/*`. The old `petbot-interactions` stack is decommissioned once (documented).
- `deploy.yml`: build + push **both** images to ECR (worker arm64, edge amd64), then apply. **Non-secret config (chat kind/model, SSM param names, region) flows from GitHub Actions Variables → `TF_VAR_*`** (the required `chat_llm_model` previously had no path into the deploy). Secret values stay in SSM.
- **`images.yml`**: build-tests both Dockerfiles on every PR (no push, no AWS creds, native arch per image).
- `ci.yml`: dropped the vestigial `--all-packages`.

**Docs**
- `deploy/terraform/README.md` rewritten (ECR-pull flow + the **complete GitHub Variables checklist** needed for a green deploy); ADR 0006 amended.

## What you must provide for a running 2.1 (merging alone won't run it)
1. One-time **bootstrap** (CloudShell): `deploy/terraform/bootstrap` → state bucket + OIDC deploy role.
2. **SSM SecureStrings**: `/petbot/edge/discord_token`; if OpenRouter, `/petbot/core/openrouter_api_key`; optional booru creds.
3. **GitHub Actions Variables**: `DEPLOY_ROLE_ARN`, `AWS_REGION`, `TF_STATE_BUCKET`, `CHAT_LLM_KIND`, `CHAT_LLM_MODEL`, (`CHAT_LLM_API_KEY_SSM_PARAMETER` for OpenRouter). Full table in the README.

Then a push to `master` runs the deploy.

## Verification (local)
- `terraform fmt -check` + `validate` (root + bootstrap) ✅
- ruff · ruff format · mypy (strict, 66) · import-linter (4/4) · pytest (58) ✅
- Docker images build-tested in `images.yml` on this PR. The first real deploy is where the Lightsail puller-role activation and ECR-pull get exercised against live AWS.

## Known follow-ups
- One-time: `terraform destroy` the old `interactions/…` state.
- App (#53, separate): `LambdaTransport` could check boto3 `FunctionError` so a Lambda-level crash payload returns a clean error instead of mis-parsing.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,10 +1,10 @@
 name: Deploy
 
-# Build + push the core worker Lambda image (arm64) and the edge container image
-# (amd64), then terraform apply — entirely from CI via GitHub OIDC (no static AWS
-# keys). See deploy/terraform/README.md. The one-time trust bootstrap (OIDC
-# provider + deploy role + state bucket + SSM secrets) is run once from AWS
-# CloudShell.
+# Build + push the core worker image (arm64) and the edge image (amd64) to ECR,
+# then terraform apply — entirely from CI via GitHub OIDC (no static AWS keys).
+# The Lightsail edge pulls its image from ECR (private-registry access). See
+# deploy/terraform/README.md. The one-time trust bootstrap (OIDC provider +
+# deploy role + state bucket + SSM secrets) is run once from AWS CloudShell.
 on:
   workflow_dispatch:
   push:
@@ -21,11 +21,18 @@ concurrency:
 
 env:
   TF_DIR: deploy/terraform
+  # Non-secret deploy config, supplied as GitHub Actions Variables and read by
+  # Terraform as TF_VAR_* (secret *values* stay in SSM, referenced by name).
+  TF_VAR_aws_region: ${{ vars.AWS_REGION }}
+  TF_VAR_chat_llm_kind: ${{ vars.CHAT_LLM_KIND }}
+  TF_VAR_chat_llm_model: ${{ vars.CHAT_LLM_MODEL }}
+  TF_VAR_chat_llm_api_key_ssm_parameter: ${{ vars.CHAT_LLM_API_KEY_SSM_PARAMETER }}
 
 jobs:
   deploy:
     # arm64 runner matches the worker Lambda's Graviton architecture, so that
-    # image builds natively; the edge image is built for amd64 via buildx.
+    # image builds natively; the smaller edge image cross-builds for amd64 via
+    # QEMU (cheaper to emulate than the heavy worker image).
     runs-on: ubuntu-24.04-arm
     # Optional manual-approval gate; configure required reviewers on the
     # `production` GitHub Environment to require sign-off before a deploy runs.
@@ -39,7 +46,7 @@ jobs:
           role-to-assume: ${{ vars.DEPLOY_ROLE_ARN }}
           aws-region: ${{ vars.AWS_REGION }}
 
-      # buildx for the gha layer cache and the amd64 edge build on an arm64 runner.
+      # buildx for the gha layer cache; QEMU for the amd64 edge build on arm.
       - uses: docker/setup-buildx-action@v3
       - uses: docker/setup-qemu-action@v3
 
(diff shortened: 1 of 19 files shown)


.github/workflows/deploy.yml
```diff
@@ -30,8 +24,8 @@ env:
 
 jobs:
   deploy:
-    # arm64 runner matches the Lambda's Graviton architecture, so the image
-    # builds natively without QEMU emulation.
+    # arm64 runner matches the worker Lambda's Graviton architecture, so that
+    # image builds natively; the edge image is built for amd64 via buildx.
     runs-on: ubuntu-24.04-arm
```
auzbuzzard, line 36: Can you explain the deploy workflow? Why are we using 1 single ubuntu arm for two build on two different platform?

auzbuzzard: Good question — and there's a deliberate reason.

**What the deploy job does (one sequential pipeline):**
1. `terraform init` + a *targeted* apply that creates just the ECR repo and the (empty) Lightsail service, then reads their names back into env vars (`EDGE_SVC`).
2. Build + **push** the worker image to ECR.
3. Build the edge image and **push** it to Lightsail; capture the returned image ref (`EDGE_IMAGE`).
4. Final `terraform apply` with both image refs → stands up the Lambda + the edge deployment.

Every step shares state with the next: the same terraform working dir/state, the OIDC AWS credentials, and values handed forward (`EDGE_SVC` → the edge push; `EDGE_IMAGE` → the final apply). That chain is why it's **one job on one runner** — splitting it across runners would mean re-authenticating OIDC per job, re-`init`-ing terraform, and passing those refs between jobs as artifacts. Lots of complexity for no real benefit on a pipeline that runs rarely.

**Why that one runner is `ubuntu-24.04-arm` (and not x86):**
The two images target different chips — worker is **arm64** (the Lambda runs on Graviton), edge is **amd64** (Lightsail container services are amd64-only). On a single runner, one of them has to cross-build under QEMU emulation. So the choice is "which one do I emulate?":
- The **worker** image is the heavy one (numpy / numexpr / pydantic-ai) → build it **natively** on arm.
- The **edge** image is tiny (discord.py + httpx + boto3, all pure-python wheels, no compilation) → emulating its amd64 build under QEMU costs ~a minute.

So: run on arm so the *expensive* image builds natively, and emulate only the *cheap* one. The `docker/setup-qemu-action` step is what enables that cross-build.

**Contrast with `images.yml` (the PR build-check):** there the two builds are *independent* — no terraform, no creds, no shared values — so I split them into two jobs on **native** runners each (arm for worker, x86 for edge). No emulation, fastest feedback. Different context → different choice.

If you'd rather avoid emulation entirely in `deploy.yml` too, the alternative is splitting it into 3 ordered stages across an arm and an x86 runner (targeted-apply → edge build/push on x86 → worker build/push + final apply), passing `EDGE_SVC`/`EDGE_IMAGE` as job outputs. It's doable but noticeably more workflow plumbing; I judged the ~1-min emulation not worth it. Happy to switch if you prefer the native-everywhere version.

---

PR #62, opened 2026-06-20
feat(edge): slash commands over the gateway (/math, /derpi, /e621)

auzbuzzard:
**Stacked on #61** (base = `claude/zen-hawking-f9db6d`; it reuses that PR's `_WORKER_UNREACHABLE`). Merge #61 first, or I can retarget to `master` after.

## Why
Slash commands don't work. RCA from the live edge logs: the Discord app still has an **Interactions Endpoint URL** set, so Discord routes interactions over HTTP (off the gateway) — discord.py warns this on every edge boot. And the code side ADR 0006 specced was never built: the edge only handled `@mention`, with no app-command handler. This PR is the code half.

## What
- **`bot.py`** — register `/math`, `/derpi`, `/e621` on the command tree, defined from the typed `*Args` surface + `Skills` client (edge still imports no skill). Sync in `setup_hook`: to `dev_guild_id` if set (instant), else global.
- **`_run_slash`** — `defer()` first (Lambda cold start can exceed Discord's 3s window), dispatch, `followup`; transport failure → friendly result.
- **`context.build_interaction_context`** — interaction twin of `build_context` (NSFW + channel id); `_channel_is_nsfw` widened to either channel type. A slash command reaches a skill identically to an `@mention`.
- **`render`** — `_plan`/`_send` factor the shaping; `respond_interaction` sends a deferred-interaction followup.

## ⚠️ Ops step required (not code, can't be done in the repo)
Clear the **Interactions Endpoint URL** in the Discord Developer Portal → General Information → Save. Until then, interactions never reach the gateway regardless of this code.

## Tests / gates
`build_interaction_context` mapping; `respond_interaction` followup; `_run_slash` defer + success + friendly-failure. ruff + format · mypy · lint-imports 4/4 · **pytest 68**.

`/music` is intentionally excluded — it lives on the separate music worker (ADR 0006 §2), not the edge.

### DIFF
--- a/src/petbot/edge/context.py
+++ b/src/petbot/edge/context.py
@@ -12,9 +12,12 @@ import discord
 from petbot.domain import Platform, SkillContext, User
 
 
-def _channel_is_nsfw(channel: object) -> bool:
-    """True if ``channel`` is an age-gated Discord channel. Typed ``object`` so it
-    serves both a message channel and an interaction channel (DMs lack the flag)."""
+def _channel_is_nsfw(
+    channel: discord.abc.MessageableChannel | discord.abc.GuildChannel | None,
+) -> bool:
+    """True if ``channel`` is an age-gated Discord channel. The union covers a
+    message's channel and a slash interaction's channel alike; ``is_nsfw`` is read
+    defensively because some channel kinds (a DM) don't carry the flag."""
     is_nsfw = getattr(channel, "is_nsfw", None)
     return bool(is_nsfw()) if callable(is_nsfw) else False
 
--- a/src/petbot/edge/context.py
+++ b/src/petbot/edge/context.py
@@ -1,7 +1,8 @@
-"""Build a neutral :class:`SkillContext` from a Discord message.
+"""Build a neutral :class:`SkillContext` from a Discord message or slash interaction.
 
 The one place Discord nouns (channels, members, NSFW flags) map onto the neutral
-context — the only direction the dependency rule allows.
+context — the only direction the dependency rule allows. An @mention and a slash
+command map through the twin functions here, so a skill sees them identically.
 """
 
 from __future__ import annotations
@@ -11,7 +12,9 @@ import discord
 from petbot.domain import Platform, SkillContext, User
 
 
-def _channel_is_nsfw(channel: discord.abc.MessageableChannel) -> bool:
+def _channel_is_nsfw(channel: object) -> bool:
+    """True if ``channel`` is an age-gated Discord channel. Typed ``object`` so it
+    serves both a message channel and an interaction channel (DMs lack the flag)."""
     is_nsfw = getattr(channel, "is_nsfw", None)
     return bool(is_nsfw()) if callable(is_nsfw) else False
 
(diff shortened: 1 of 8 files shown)


src/petbot/edge/context.py
```diff
@@ -11,7 +12,9 @@
 from petbot.domain import Platform, SkillContext, User
 
 
-def _channel_is_nsfw(channel: discord.abc.MessageableChannel) -> bool:
+def _channel_is_nsfw(channel: object) -> bool:
```
auzbuzzard: Moving from interaction to edge should not require erasing types.

src/petbot/edge/render.py
```diff
@@ -31,29 +35,49 @@ def to_embed(spec: EmbedSpec) -> discord.Embed:
...
-        return
-
-    for index, chunk in enumerate(chunks):
-        # The embed rides only the first message; pass it only when present so the
-        # call matches discord.py's non-optional `embed` overload.
-        if embed is not None and index == 0:
-            await channel.send(content=chunk, embed=embed)
-        else:
-            await channel.send(content=chunk)
+async def respond_interaction(interaction: discord.Interaction, result: SkillResult) -> None:
```
auzbuzzard, line 76: why do we need to send? Explain, they're calling the same Skills?

src/petbot/skills/chat/agent.py
```diff
@@ -44,34 +45,34 @@ def _summarise(deps: ChatDeps, result: SkillResult) -> str:
     return "\n".join(parts) or "Done."
 
 
+def _tool_for(spec: CommandSpec) -> Tool[ChatDeps]:
```
auzbuzzard, line 48: Why do you need to modify the Chat Skills for this?

auzbuzzard (commented):
With Interaction, slash commands were registered with just a manifest? Why does it require that many changes in the edge dispatcher in order to register commands?

src/petbot/edge/context.py
auzbuzzard: Fixed in 9f406a9. You're right — I only widened it to `object` to share the helper across a message channel and an interaction channel, which threw the type away. It's now `discord.abc.MessageableChannel | discord.abc.GuildChannel | None` (the union both call sites actually pass). The `getattr` guard stays only because some channel kinds — a DM — genuinely don't carry `is_nsfw`; the type is no longer erased.

src/petbot/edge/render.py
auzbuzzard: The Skills call and the resulting `SkillResult` are identical — that's the shared dispatch. What differs is only how Discord wants the **reply** delivered, and it forces two primitives:

- **@mention**: a normal message arrived, so we reply by posting into the channel (`channel.send`).
- **slash**: an *interaction* arrived, and Discord requires the reply go back through the interaction (`interaction.response` / `followup.send`) within ~3s or it shows "application did not respond". Posting to the channel would leave the interaction unacknowledged.

So the only real difference is the final sender callable. That's why the shaping is shared — `_plan` builds the identical message list — and `_send` takes the sender; `respond` / `respond_interaction` are 2-line wrappers picking `channel.send` vs `interaction.followup.send`. If you'd rather not have two named wrappers, I can collapse them into one `deliver(sender, result)`.

src/petbot/skills/chat/agent.py
auzbuzzard: Slash doesn't need it — the edge works without touching the agent. I changed it so the manifest is a **single** source: if I'd left the agent's hand-written `@agent.tool` trio, the new `COMMANDS` list would just be the edge's private copy and the agent would remain a second list — the exact duplication this refactor set out to kill. Looping the agent over the same manifest is what makes it one declaration → two consumers (LLM tools + slash commands).

It's separable, though. If you'd prefer #62 stay edge-only, I'll revert the agent change here and land the agent→manifest unification as its own PR (the manifest stays in `petbot.types`; only the edge consumes it for now). Your call.

auzbuzzard:
Closing as superseded by #63, which reimplements this slash-command work on the shared command pipeline (typed catalog, injected I/O ports, public `app_commands.describe` instead of the private attribute). #63 is based directly on `master` and is independent of #61, so this PR is no longer needed.

---

PR #64, opened 2026-06-21
Conversational memory: reply-to-continue + reactive context handling (ADR 0010)

auzbuzzard:
## What

Makes PetBot multi-turn: **reply to a PetBot message and it continues the conversation** — no re-@mention needed. Previously chat was stateless single-shot (an `@mention` ran the agent with no `message_history`).

Built on the post-#66 process pipeline (this branch was rebased onto current `master`; the earlier pre-refactor commit was dropped).

## How

**Memory — reply-to-continue, reconstructed frontend-side (Route A).**
- `frontends/discord/history.py` walks the `message.reference` chain (bounded by `HISTORY_MAX_TURNS`, default 25) into neutral `Turn` values.
- History rides on **`TextInput.history`** — the conversational variant of the `Input` sum type, so a `CommandInput` can't carry chat history; not on the shared `SkillContext`, and not a compute-side store (which can't see Discord's reply chain). The compute service stays a stateless, pure function of `(Input, SkillContext)`.
- `Role` is neutral `USER`/`ASSISTANT`; the name "PetBot" lives in `Turn.author` (from the live Discord display name). A user turn inlines its author for multi-user attribution; an **image-only bot card is flattened to a faithful text description** (title + image URL), since an assistant turn is text-only in the model's history.
- PetBot answers as a Discord reply (`mention_author=False`) so each reply is an anchor.

**Context window — reactive, no magic number.**
- pydantic-ai doesn't expose a model's context-window size (upstream [#4538](https://redirect.github.com/pydantic/pydantic-ai/issues/4538)) and has no turnkey compaction ([#4137](https://redirect.github.com/pydantic/pydantic-ai/issues/4137)), so a configured token budget would be wrong on a model swap.
- Instead the **provider is the trigger**: the chat process attempts the run, and only if it's rejected for length does it compact and retry (bounded). Strategy is **DI-selectable** via `CHAT_CONTEXT__KIND` (a config `match`, like `build_model_from_config`): a zero-cost **sliding window** or an **LLM summarizer** (stylizer tier). `to_model_messages` stays a pure mapper.

## DDD/DI notes
- Memory enters the pipeline at one principled point (`TextInput.history`); the core stays pure.
- The compaction strategy mirrors the existing `StylePort`/`Stylist` + `build_model_from_config` patterns; it's process-level (operates on pydantic-ai `ModelMessage`), not a domain port. The only added conditionals are exhaustive `match`es on closed types + one error boundary.

## Tests & gates
`102 passed`; `ruff` / `ruff format` / `mypy --strict` / `lint-imports` all green. New/updated tests cover the mapper, both compaction strategies, overflow detection + the chat retry path, reply-chain reconstruction (incl. image-card flattening and graceful stops), reply threading, and the `Input` round-trip.

See [`docs/adr/0010-conversational-memory-reply-to-continue.md`](docs/adr/0010-conversational-memory-reply-to-continue.md).

## Out of scope
Compute-side `SessionStore`; reply-about-another-message as explicit quoting; proactive precise token budgeting (deferred until the upstream window-size gap is fixed); the `Notifier`/streaming path.

https://claude.ai/code/session_01CV2b3SvKc1qptAC4G7NTTv

### DIFF
--- a/src/petbot/domain/input.py
+++ b/src/petbot/domain/input.py
@@ -14,6 +14,7 @@ re-hydrate the right member from a wire payload.
 
 from __future__ import annotations
 
+from enum import StrEnum
 from typing import Annotated, Literal
 
 from pydantic import Field
@@ -21,11 +22,42 @@ from pydantic import Field
 from petbot.domain._model import Frozen
 
 
+class Role(StrEnum):
+    """Who spoke a :class:`Turn` — the structural position in the conversation.
+
+    Neutral and platform-agnostic (it maps onto the user/assistant split every chat
+    model uses); the speaker's *name* lives on :attr:`Turn.author`, not here.
+    """
+
+    USER = "user"
+    ASSISTANT = "assistant"
+
+
+class Turn(Frozen):
+    """One prior message in the conversation a :class:`TextInput` continues.
+
+    ``author`` is the speaker's display name (a user's, or — for an assistant turn —
+    PetBot's own); ``text`` is what was said. An image-only bot reply is flattened to a
+    faithful text description by the frontend, so ``text`` is never structured data.
+    """
+
+    role: Role
+    author: str
+    text: str
+
+
 class TextInput(Frozen):
-    """Free-text conversational input (an ``@mention``). The process interprets it."""
+    """Free-text conversational input (an ``@mention``). The process interprets it.
+
+    ``history`` is the prior turns of the conversation this message continues (the
+    reply chain, oldest-first), reconstructed by the frontend; empty for a fresh
+    @mention. It rides on the conversational variant — never on ``CommandInput`` — so a
+    command can't carry chat history.
+    """
 
     kind: Literal["text"] = "text"
     text: str
+    history: tuple[Turn, ...] = ()
 
 
 class CommandInput(Frozen):
--- a/src/petbot/frontends/discord/history.py
+++ b/src/petbot/frontends/discord/history.py
@@ -1,0 +1,119 @@
+"""Reconstruct a conversation's history from the Discord reply chain.
+
+The frontend is the only place that holds the gateway, so it is the only place that can
+walk ``message.reference`` to recover the turns a reply continues. This maps that chain
+onto neutral :class:`~petbot.domain.input.Turn` values for ``TextInput.history`` — the
+driving-adapter half of "map a platform event to a complete ``Input``".
+
+Two parts: an async :func:`walk_reply_chain` that does the Discord I/O (bounded, and
+tolerant of deleted/unreadable ancestors), and a pure :func:`to_turns` that maps the
+gathered raw turns to neutral ones. The history is shipped *faithful and untrimmed* — an
+over-long conversation is handled compute-side (reactively), not here. An image-only bot
+card is flattened to a faithful text description (its real title + image URL), since an
+assistant turn can only be text in the model's history.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from dataclasses import dataclass
+
+import discord
+
+from petbot.domain import Role, Turn
+
+
+@dataclass(frozen=True, slots=True)
+class RawTurn:
+    """One ancestor message, read off Discord, before neutral mapping."""
+
+    display_name: str
+    is_self: bool
+    text: str
+
+
+def _describe_card(embed: discord.Embed) -> str:
+    """A faithful one-line description of an embed (title + image URL), or ``""``."""
+    parts = [part for part in (embed.title, embed.image.url) if part]
+    return f"[image] {' — '.join(parts)}" if parts else ""
+
+
+def _message_text(message: discord.Message) -> str:
+    """A message's text: its content, else a description of its first usable embed."""
+    if message.content:
+        return message.content
+    for embed in message.embeds:
+        described = _describe_card(embed)
+        if described:
+            return described
+    return ""
+
+
+async def walk_reply_chain(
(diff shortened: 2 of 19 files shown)


src/petbot/domain/input.py
```diff
@@ -14,18 +14,50 @@
...
+    """Free-text conversational input (an ``@mention``). The process interprets it.
+
+    ``history`` is the prior turns of the conversation this message continues (the
+    reply chain, oldest-first), reconstructed by the frontend; empty for a fresh
+    @mention. It rides on the conversational variant — never on ``CommandInput`` — so a
+    command can't carry chat history.
+    """
 
     kind: Literal["text"] = "text"
     text: str
```
auzbuzzard, line 59: Would text benefit from being upgraded to a Turn()? Consequences? I feel it will make things consistent, but I want to know possible side effects.

src/petbot/frontends/discord/history.py
```diff
@@ -0,0 +1,129 @@
...
+from collections.abc import Callable
+from dataclasses import dataclass
+
+import discord
+
+from petbot.domain import Role, Turn
+
+
+@dataclass(frozen=True, slots=True)
+class RawTurn:
```
auzbuzzard: Why do we need a RawTurn? Do we not already have a Turn model? Discuss.

src/petbot/frontends/discord/settings.py
```diff
@@ -46,6 +46,9 @@ class DiscordSettings(BaseSettings):
     discord_token: str
     dev_guild_id: int | None = None
     log_level: str = "INFO"
+    #: How many reply-chain ancestors to fetch when reconstructing conversation history
+    #: (a Discord-API cost bound, not a model-context bound). Env ``HISTORY_MAX_TURNS``.
+    history_max_turns: int = 25
```
auzbuzzard, line 51: No Magic number. Put it on Env var with documentation.

src/petbot/domain/input.py
auzbuzzard: Recommend keeping `text: str`.

A `Turn` carries `role` + `author` because *history* has many authors and both roles. The current message has neither variation:
- `role` is always `USER` — a `Turn` makes `role=ASSISTANT` representable (an illegal state the sum-type design otherwise forbids).
- `author` is always the requester, which already lives on `SkillContext.user.display_name`, so upgrading duplicates it (drift risk). History's `Turn.author` *isn't* redundant because those authors vary; the current one doesn't.

No consumer gain either: pydantic-ai splits `run(user_prompt: str, message_history=[...])`, so `ChatProcess` would just read `.text` for `user_prompt`, and the current message is never author-inlined (it's the live speaker) unlike history turns.

So the `str` / `tuple[Turn, ...]` split is the honest model — it mirrors pydantic-ai's own prompt-vs-history split — not an inconsistency. If reader clarity is the worry, a one-line doc on `text` is the cheap fix (no type change).

---

src/petbot/frontends/discord/settings.py
auzbuzzard: This is already env-configurable + documented — line 51 on its own hides that:

- **Env var:** `DiscordSettings` sets no `env_prefix`, so the field maps straight to `HISTORY_MAX_TURNS` (same as `DISCORD_TOKEN`). Already overridable.
- **Docs:** the field docstring (lines 49–50) and `.env.example:32–34` (`# HISTORY_MAX_TURNS=25`).

On "magic number": unlike the context-window budget — which we made *reactive* precisely because there's no honest per-model default — this `25` is a **cost knob** (how many REST hops to spend walking the reply chain). It degrades gracefully and legitimately has a sane default, so a documented + env-overridable default is the right shape, not a hidden constant.

Happy to make it **required** (no default), forcing every deploy to set it, if you'd prefer — though I'd advise against for a tuning knob. Otherwise resolving this.

---

src/petbot/frontends/discord/history.py
auzbuzzard: `RawTurn` and `Turn` sit in **different layers with different fields**:

- `RawTurn{display_name, is_self, text}` — Discord-side, raw (pre-strip, may be empty)
- `Turn{role, author, text}` — neutral, clean (mention-stripped, non-empty)

`RawTurn` is the I/O-boundary projection: the async `walk_reply_chain` just reads messages off Discord into `RawTurn`s, and the **pure** `to_turns` does the neutral mapping (`is_self`→`role`, strip mention, drop empties, reverse to oldest-first). That seam is what lets `to_turns` be a pure, `discord`-free function tested with trivial 3-field fixtures.

Dropping it (walk returns `list[discord.Message]`, map straight to `Turn`) would couple `to_turns` to `discord.Message` and its tests to full message fakes — a worse trade. So I'd **keep the type**.

Agreed the *name* is weak, though. Proposing **`DiscordTurn`** — the Discord-layer counterpart of the neutral `Turn`, which makes the two-layer mapping self-documenting (`to_turns: list[DiscordTurn] -> tuple[Turn, ...]`). Alternative if you'd rather not have two `*Turn` types: `ChainMessage`. Your pick.

---

auzbuzzard: Use DiscordTurn.

auzbuzzard: Done — renamed `RawTurn` → `DiscordTurn` (type + `walk_reply_chain`/`to_turns`/`reconstruct` signatures + test fixtures + docstrings) in `30a1f32`. Pure rename, no behaviour change; ruff/mypy/lint-imports/tests green.

---

PR #74, opened 2026-06-24
feat(observability): metadata-only OpenTelemetry agent observability (ADR 0011)

auzbuzzard:
## Why

An incident — PetBot silently not calling the `e621` skill — was **undiagnosable from logs**. The core Lambda ran clean (zero warnings/errors over two days) because the failure is model *behaviour*: the agent declined the tool, or the search came back empty, and neither path logged anything. PetBot recorded none of the LLM-agent essentials (tool-call decisions, tokens, latency, a correlation id), and the Lambda configured **no app logging at all** — every INFO record dropped at the root WARNING default.

## What

Wire pydantic-ai's first-class OpenTelemetry instrumentation (already on 1.x, unused) plus a few non-content signals, exported AWS-native (X-Ray + CloudWatch EMF) over OTLP. **Off by default** (`OBS_ENABLED`). See [ADR 0011](docs/adr/0011-agent-observability.md) and the [telemetry schema](docs/telemetry.md).

- **`petbot/observability.py`** — `configure_observability` / `flush_observability` / `hash_user_id`, wired once per entrypoint like `configure_logging`.
- **Agent instrumented** with `include_content=False` (metadata only). `ChatProcess` emits one run-outcome log per turn (tools, tokens, finish reason, recalled/unrecalled context, salted-hashed user id) — the always-on CloudWatch path that survives the collector being down — plus counters (tool-calls, zero-tool, lost-context).
- **Booru outcome** (`ok|empty|safe_limited|error`) on the tool span — recovers "why nothing came back" without storing tags.
- **Distributed trace** `dispatch → serve → agent → tool` via a W3C carrier on the `Dispatch` wire. **Domain kernel untouched** (`SkillContext` unchanged — trace context is transport metadata).
- **Lambda handler** now configures structured JSON logging + telemetry at cold start and flushes spans per invocation (fixes the original "no app logs" blind spot).
- **Terraform** — gated (default-off) env + IAM plumbing for X-Ray / CloudWatch.

## Privacy & licensing

- **Metadata-only**: no prompts, tags, or replies are ever recorded; the one identifier is a salted hash. `PRIVACY.md` (hosted-service; Discord Message-Content intent + GDPR) documents it; per-guild consent is deliberately not needed (data minimisation).
- **Apache-2.0** `LICENSE` + `NOTICE` (pyproject declared MIT but shipped no license file).

## Dependencies

Raised every floor to its locked minor — `pydantic-ai-slim>=1.107` (hard requirement for `InstrumentationSettings`) plus hygiene on the rest.

## Deploy notes (not auto-validated)

- Terraform is **default-off** (zero risk to current deploys); the collector itself is the remaining step — the core is a **container-image** Lambda, so the ADOT *layer* doesn't apply (bake the collector extension into the image, or point at the edge collector). No `terraform` locally → needs a real `plan`.
- `PRIVACY.md` has placeholders (controller name, contact, jurisdiction); `LICENSE` copyright holder defaulted to `auzbuzzard`.

## Testing

`ruff` / `ruff format` / `mypy` / `lint-imports` (6/6 contracts) clean; **129 tests** — incl. no-content-leak in spans, trace re-parenting, run-outcome metadata-only, booru outcome classification, recalled/unrecalled context. A `pydantic-evals` harness guards e621 tool selection (real-model, manual/scheduled).


src/petbot/skills/booru/skill.py
```diff
@@ -25,6 +26,21 @@
...
+_OUTCOMES = _meter.create_counter(
+    "petbot.booru.outcome", description="Booru searches by provider + outcome status."
+)
+
+
+def _record_outcome(provider: str, status: str) -> None:
+    """Tag the active tool span with the coarse outcome and bump the counter.
+    ``status`` ∈ ``ok | empty | safe_limited | error``."""
+    trace.get_current_span().set_attribute("petbot.booru.outcome", status)
+    _OUTCOMES.add(1, {"provider": provider, "status": status})
```
auzbuzzard: Shouldn't the outcome status string be defined by the Exception? Now skill is holding the definition of the Exception. Or at least, the str needs to be Literallized somewhere.

src/petbot/services/core/handler.py
```diff
@@ -46,6 +61,11 @@ def handler(event: dict[str, Any], context: object = None) -> dict[str, Any]:
     itself, which the frontend parses directly.
     """
     body = event["body"] if isinstance(event, dict) and "body" in event else json.dumps(event)
```
auzbuzzard, line 62: is isinstance(event, dict) necessary here?

auzbuzzard (commented):
1. Review all the comments. They should be time-invariant and not dependent on the PR. They should also not be over. verbose.
2. Why is there a top level eval added? Is this part of test? Can we not add top level dir without concent?

src/petbot/skills/booru/skill.py
auzbuzzard: Centralized the statuses as a `BooruOutcome` StrEnum in `skills/booru/types.py` (af083da) — no more loose literals.

Kept the classification in the skill rather than on the exceptions, deliberately: `SAFE_LIMITED` vs `EMPTY` depends on the request's `safe_only` floor (not knowable from the exception type alone), and the empty case raises a *domain* `EmptyResult` that shouldn't carry booru-specific vocabulary. So the skill maps `(exception + request) -> BooruOutcome`, and the enum is now the single source of the strings.

src/petbot/services/core/handler.py
auzbuzzard: That line is pre-existing — not part of this PR (I only wrapped the call below in `try/finally` for the span flush). Its job is to accept both a direct boto3 invoke (`{"input","context"}`) and an API-Gateway-style `{"body": ...}` wrapper.

The `isinstance(event, dict)` guards the `"body" in event` test against a non-dict `event`: a misconfigured invoke can pass a JSON list/scalar, where `in` would `TypeError`. Given the `event: dict[str, Any]` annotation it is belt-and-suspenders. Happy to drop it in a separate cleanup, but left it untouched to keep this PR scoped to observability.

auzbuzzard:
Addressed both points in af083da:

1. **Time-invariant comments** — scrubbed incident/PR-dependent phrasing from code comments and docstrings (e.g. "the e621 blind spot", "the original blind spot", "the exact e621 failure mode") and trimmed the over-verbose ones; they now describe behavior, not history. (ADR 0011's *Context* section still records the incident — that's the ADR's purpose, not source comments.)

2. **Top-level `evals/`** — removed it, plus the now-orphan `eval` extra. You're right that a top-level dir shouldn't be added unprompted. It wasn't part of the offline suite: `TestModel` can't judge tool *selection*, so it was a manual/scheduled real-model harness, not a test. If you want eval coverage later, I'll add it under an agreed location.

